PR #1, opened 2022-02-06
logger service

jjalan:
## Description
Logger service - log to console for local and testing env, log to rollbar for staging, qa, beta and production.

## Database schema changes
`None`

- Loggers should initialize

### Manual test cases run
- Log to rollbar
- Log to console

### DIFF
--- a/src/modules/logger/internals/console-logger.ts
+++ b/src/modules/logger/internals/console-logger.ts
@@ -1,6 +1,6 @@
 import Logger from './types';
 
-export default class ConsoleLogger implements Logger {
+export class ConsoleLogger implements Logger {
   // eslint-disable-next-line class-methods-use-this
   public info(message: string): void {
     // eslint-disable-next-line no-console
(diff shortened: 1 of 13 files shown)


src/modules/logger/internals/rollbar-logger.ts
```diff
@@ -0,0 +1,34 @@
...
+  public warn(message: string): void {
+    this.rollbar.warning(message);
+  }
+
+  public criticial(message: string): void {
+    this.rollbar.critical(message);
+  }
+}
+
+const rollbarLogger = new RollbarLogger();
```
jjalan, line 32: We should let the consumer of this class to instantiate.

src/modules/logger/internals/console-logger.ts
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  public criticial(message: string): void {
+    // eslint-disable-next-line no-console
+    console.log(`[CRITICAL] ${message}`);
+  }
+}
+
+const consoleLogger = new ConsoleLogger();
```
jjalan, line 29: Let consumer of this class instantiate.

jjalan (commented):
LGTM, after addressing comments

PR #2, opened 2022-02-06
config service

jjalan:
## Description
Added config service to get env variables.

## Database schema changes
`None`

- ConfigService should return the correct environment
- ConfigService should return correct env variable value
- ConfigService should throw MissingKeyError if the key is not found in the config
- ConfigService should throw a ValueTypeMismatchError if type of value does not match


package.json
```diff
@@ -9,8 +9,9 @@
   "main": "src/main.ts",
   "scripts": {
     "build": "gulp",
-    "test": "exit 0",
+    "test": "NODE_ENV=testing mocha --exit -r ts-node/register --project tsconfig.json test/**/*.spec.ts",
```
jjalan, line 12: NODE_CONFIG_ENV (I think) should be testing and NODE_ENV as production.

This instructs node to compile libs in production mode but use testing config.

src/modules/config/config-service.ts
```diff
@@ -0,0 +1,70 @@
...
+import {
+  ConfigType,
+  Environment,
+  MissingKeyError,
+  ValueTypeMismatchError,
+} from './types';
+
+export default class ConfigService {
+  public static getEnvironment(): Environment {
+    return config.get('node.env');
```
jjalan, line 11: Please read and map to enum. If it cannot be mapped to enu, it should throw

src/modules/config/config-service.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  public static getBoolValue(key: string): boolean {
+    // This method should throw MissingKeyError if key is not found in config
+    // This method should throw ValueTypeMismatchError if key is found but the
+    // there is a value type mistmatch
+    // Otherwise return the value
+    const expectedValueType = ConfigType.BOOLEAN;
+    const value = config.get(key);
```
jjalan, line 20: I see this code repeated below.

Possible to refactor?

config/custom-environment-variables.json
```diff
@@ -1 +1,5 @@
-{}
+{
+  "node": {
+    "env": "NODE_ENV"
```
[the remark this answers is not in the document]
jjalan: Can I request to reconfirm this by reading config npm module documentation?

jjalan (commented):
We should also add a file `qa.json` and `beta.json`.

config/custom-environment-variables.json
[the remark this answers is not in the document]
jjalan: We should add a spec to read a sample value from each environment.

src/modules/config/config-service.ts
```diff
@@ -0,0 +1,51 @@
+import config from 'config';
+import { Environment, MissingKeyError, ValueTypeMismatchError } from './types';
+
+export enum ConfigType {
```
jjalan, line 4: Please move it to internals/config-type.ts

Service should not contain details or should contain very minimal detail so reader of this file may not know much about implementation details please.

src/modules/config/config-service.ts
```diff
@@ -0,0 +1,51 @@
...
+  STRING = 'string',
+}
+
+export default class ConfigService {
+  public static getEnvironment(): Environment {
+    const env: Environment = config.get('node.config.env');
+    if (Object.values(Environment).includes(env)) {
+      return env;
+    }
+    throw new Error(`Environment ${env} is not supported`);
```
jjalan, line 16: We should create `EnvironmentNotSupportedError`

src/modules/config/config-service.ts
```diff
@@ -0,0 +1,51 @@
...
+    return this.getEnvValue<number>(expectedValueType, key);
+  }
+
+  public static getStringValue(key: string): string {
+    const expectedValueType = ConfigType.STRING;
+    return this.getEnvValue<string>(expectedValueType, key);
+  }
+
+  private static getEnvValue<T>(expectedValueType: ConfigType, key: string): T {
+    // This method should throw MissingKeyError if key is not found in config
```
jjalan, line 35: Remove comments, code is readable.

test/spec/config/config.spec.ts
```diff
@@ -0,0 +1,58 @@
+/* eslint-disable @typescript-eslint/no-unsafe-call */
+import { assert } from 'chai';
+import { Environment } from '../../../src/modules/config/types';
+import ConfigService from '../../../src/modules/config/config-service';
+
+describe('ConfigService', () => {
```
jjalan, line 6: consider adding spec for all environments please.

akashlakhera: It's very hard to change the config at runtime and make the app throw an `EnvironmentNotSupportedError` error.

jjalan (commented):
LGTM

PR #8, opened 2022-02-07
 Email service, SMS service integrated with config service. 

Sparsh-Kumar:
## Description
- Completed Email service of communication service and it's test cases.
- Added test case to ensure that If provide invalid email (recipient, sender) then email service will throw error.
- Added test case to ensure that if provide empty sender name , then email service will throw error.
- Added test case to ensure that if all details provided(recipient email, sender email, sender name) are valid, then email service will send an email.
- Completed Sms service of communication service and it's test cases.
- Added test case to ensure to not allow sending sms, if recipient phone number is invalid.
- Added test case to ensure to not allow sending sms, if sms body is empty.

## Database schema changes
- N/A

- should not allow to send email, if recipient email is invalid.
- should not allow to send email, if sender email is invalid.
- should not allow to send email, if sender name is empty.
- should allow to send email,  if recipient email, sender email, sender name are valid.
- should not allow to send sms, if the reciepient phone number is invalid
- should not allow to send sms, if the message body is empty.

### Manual test cases run
- N/A


src/modules/communication/email-service.ts
```diff
@@ -0,0 +1,10 @@
+import SendGridService from './internals/send-grid-service';
+import { SendEmailParams } from './types';
+import { SendGridEmailParams } from './internals/sendgrid-email-params';
+export default class EmailService {
+  public static async sendEmail(params: SendEmailParams): Promise<void> {
+    const sendGridParams = new SendGridEmailParams(params);
```
jjalan, line 6: Validation should happen inside `SendGridService.sendEmail`. The root level service should be implementation free

src/modules/communication/index.ts
```diff
@@ -0,0 +1,11 @@
+import EmailService from './email-service';
+import SMSService from './sms-service';
+import { SendGridValidationError } from './types';
+import SendGridService from './internals/send-grid-service';
+
+export {
+  EmailService,
+  SMSService,
+  SendGridValidationError,
```
jjalan, line 9: Nothing about sendgrid should be visible to outside world from here

src/modules/communication/internals/send-grid-service.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export default class SendGridService {
+  public static initializeService() {
+    const sendgridAPIKey:string = config.get('sendgridAPIKey'); // Todo: here use configuration service to get these details.
+    if(sendgridAPIKey) {
+      mail.setApiKey(sendgridAPIKey);
+    }
+  }
+  public static async sendEmail(params: SendEmailParams): Promise<void> {
+    const msg: any = {
```
jjalan, line 13: Do not use `any`

src/modules/communication/internals/send-grid-service.ts
```diff
@@ -0,0 +1,32 @@
...
+    msg.templateId = params.templateId;
+    msg.dynamic_template_data = params.templateData;
+    msg.dynamic_template_data = Object.assign(msg.dynamic_template_data, {
+      domain: config.get('webAppHost'),
+    });
+    try {
+      await mail.send(msg);
+    } catch(error) {
+      // Todo: Use logger Service to log these values
+      console.log(error.message);
```
jjalan, line 29: We should throw error so the caller knows something went wrong and can take corrective action.

We should create ThirdPartyServiceError and throw its instance from here

jjalan (changes requested):

src/modules/communication/internals/sendgrid-email-params.ts
```diff
@@ -0,0 +1,59 @@
+import {
+  LooseObject,
+  SendEmailParams,
+  SendGridValidationError,
+  ValidationFailure
+} from "../types";
+
+export class SendGridEmailParams {
```
jjalan, line 8: This is not specific to send grid I believe, best to rename it to SendEmailPArams

src/modules/communication/internals/sendgrid-email-params.ts
```diff
@@ -0,0 +1,59 @@
...
+        })
+      }
+      if(!isfromNameValid) {
+        failures.push({
+          field: 'fromName',
+          message: 'Please specify a valid name of the sending recipient.',
+        }); 
+      }
+      throw new SendGridValidationError(
+        'Cannot send email, as one or more params are invalid.',
```
jjalan, line 48: I have requested in the past on how to write good error message. Please Google about it

test/specs/sendgrid.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+  });
+  it('should send email, if all the values are valid.', async function() {
+    const params = {
+      to: 'to@gmail.com',
+      from: 'from@gmail.com',
+      fromName: 'Sample Name',
+      templateId: 'd-bea6fc9b30d34caeb73c11e2770e3287',
+      templateData: {}
+    };
+    sinonSandbox.stub(SendGridService, <any>'sendEmail').returns(Promise.resolve('1'));
```
jjalan, line 90: Poor code spacing. Need line breaks

src/modules/communication/internals/send-grid-service.ts
```diff
@@ -0,0 +1,32 @@
+import { SendEmailParams } from '../types';
+import config from 'config';
+import mail from '@sendgrid/mail';
+
+export default class SendGridService {
+  public static initializeService() {
+    const sendgridAPIKey:string = config.get('sendgridAPIKey'); // Todo: here use configuration service to get these details.
+    if(sendgridAPIKey) {
```
jjalan, line 8: This does not sound right, if the caller has asked SendGridService.initializeService and if config is not, it should throw.

Failing silently results in unexpected behaviours.

src/modules/communication/internals/sendgrid-email-params.ts
```diff
@@ -0,0 +1,59 @@
...
+    this.templateId = params.templateId;
+    this.templateData = params.templateData;
+  }
+
+  async validate() {
+    const failures: ValidationFailure[] = [];
+    const isToValid = this.isEmailValid(this.to);
+    const isFromValid = this.isEmailValid(this.from);
+    const isfromNameValid = !!this.fromName;
+    if(!isToValid || !isFromValid || !isfromNameValid) {
```
jjalan, line 28: Why is this needed?

Can we simply not check each ifs and push errors.

src/modules/communication/internals/sendgrid-email-params.ts
```diff
@@ -0,0 +1,59 @@
...
+      }
+      throw new SendGridValidationError(
+        'Cannot send email, as one or more params are invalid.',
+        failures
+      );
+    }
+  }
+
+  isEmailValid(email: string) {
+    const re =
```
jjalan, line 55: Create a constants file and store the regex there.

src/modules/communication/index.ts
Sparsh-Kumar: @jjalan  sir, SendGridValidationError, is just another class initiated by me, it is not from sendgrid. This class is defined in types.ts.

jjalan: Understood, it cannot be in index.ts file for the module.

Sendgrid is an internal detail

src/modules/communication/internals/sendgrid-email-params.ts
```diff
@@ -0,0 +1,59 @@
...
+          message: 'Please specify a valid from email.'
+        })
+      }
+      if(!isfromNameValid) {
+        failures.push({
+          field: 'fromName',
+          message: 'Please specify a valid name of the sending recipient.',
+        }); 
+      }
+      throw new SendGridValidationError(
```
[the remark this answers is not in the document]
Sparsh-Kumar: the ```sendGridValidationError``` is just a class defined by me, It is not related to sendgrid , we are not throwing anything related to sendgrid to client.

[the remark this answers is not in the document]
jjalan: Who will catch this error?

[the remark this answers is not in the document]
Sparsh-Kumar: @jjalan  sir, actually I think the sendEmail service would be used by other service, so any controller which would make instance of this sendgrid-email-params, would contain a try catch to catch this error.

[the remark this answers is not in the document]
Sparsh-Kumar: ok, @jjalan  sir, got it, I will fix that, please also let me know if you have any workaround for this.

src/modules/communication/constant.ts
```diff
@@ -0,0 +1 @@
+export const emailRegex = /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
```
jjalan, line 1: This is an internal detail, should be inside internal folder

src/modules/communication/index.ts
```diff
@@ -0,0 +1,9 @@
+import EmailService from './email-service';
+import SMSService from './sms-service';
+import SendGridService from './internals/send-grid-service';
+
+export {
+  EmailService,
+  SMSService,
+  SendGridService,
```
jjalan, line 8: SendGridService should not be exported

src/modules/communication/types.ts
```diff
@@ -0,0 +1,57 @@
...
+  to: string;
+};
+
+export type SendSMSParams = {
+  to: PhoneNumber,
+  message: string
+};
+
+export enum ErrorCodes {
+  VALIDATION_ERROR = 404,
```
jjalan, line 24: Please follow other module convention for error codes

src/modules/communication/types.ts
```diff
@@ -0,0 +1,57 @@
...
+  [key: string]: any;
+};
+
+export type PhoneNumber = {
+  countryCode: string;
+  phoneNumber: string
+};
+
+export type SendEmailParams = {
+  from: string;
```
jjalan, line 11: Best to combine `from`, `fromName` into 
```
export type EmailSender {
  email: string;
  name: string;
}
```
Same thing for to
```
export type EmailRecipient {
  email: string;
}
```

src/modules/communication/types.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export class MissingConfigVariable extends Error {
+  constructor(variableName: string) {
+    super(`Missing config variable ${variableName}.`);
+  }
+};
+
+export class ThirdPartyServiceError extends Error {
+  code: ErrorCodes;
+  constructor() {
```
jjalan, line 53: you should pass third party name and error message from it in constructor. The error should be something like:

`${thirdParty} encountered error: ${thirdPartyErrorMessage}`

test/spec/email/sendgrid.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+  EmailService,
+} from '../../../src/modules/communication';
+import { initApp } from '../../helpers/spec-helpers';
+import mail from '@sendgrid/mail';
+
+chai.use(chaiHttp);
+let expressApp: Application;
+let sinonSandbox: sinon.SinonSandbox;
+
+describe('Email Service.', function() {
```
jjalan, line 16: Specs are hard to read, there are no logical line breaks.

Ex: See this - https://github.com/sendgrid/sendgrid-nodejs/blob/78dae9ed05cc54bd50b4c0354a5e15855d83db60/packages/helpers/classes/response-error.js

You will see how there is logical chunk of separation.

PR #11, opened 2022-02-11
access token params

akashlakhera:
## Description
Update access token params

## Database schema changes
`None`

- POST /access-tokens should return access token for given username password

### Manual test cases run
- User should be able to log in

### DIFF
--- a/src/modules/access-token/access-token-service.ts
+++ b/src/modules/access-token/access-token-service.ts
@@ -1,4 +1,3 @@
-import AccountService from '../account/account-service';
 import AccessTokenWriter from './internal/access-token-writer';
 import { AccessToken, CreateAccessTokenParams } from './types';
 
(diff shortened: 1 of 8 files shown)


src/modules/access-token/access-token-service.ts
```diff
@@ -1,10 +1,18 @@
...
-    return AccessTokenWriter.createAccessToken(params);
+  ): Promise<AccessToken> {
+    const accountSearchParams = {
+      username: params.username,
+      password: params.password,
+    };
+    const account = await AccountService.getAccountByUsernamePassword(
+      accountSearchParams,
+    );
+    return AccessTokenWriter.createAccessToken(account.id);
```
jjalan, line 16: Move line 10 - 16 inside `AccessTokenWriter.createAccessToken`. Remember, keep service file light weight

test/spec/access-token/access-token.spec.ts
```diff
@@ -32,17 +33,16 @@ describe('POST /access-tokens', () => {
     sinonSandbox.restore();
   });
 
-  it('should return access token for given account id', async () => {
+  it('should return access token for given username password', async () => {
     sinonSandbox.stub(ConfigService, 'getStringValue').returns('1h');
 
-    const accountId = 'testAccountId';
+    const params = { username: 'username', password: 'password' };
```
jjalan, line 39: Don't you need to create an account first?

PR #15, opened 2022-02-17
feat: Adding param validators to account api

vindru:
## Description

Adding param validators to account api

## Database schema changes
NA

## Test cases

- POST /account should throw validation error if username is invalid
- POST /account should throw validation error if password is weak


src/modules/account/internal/account-param-validation.ts
```diff
@@ -0,0 +1,55 @@
...
+        failures,
+      );
+    }
+  }
+
+  public static isUsernameValid(
+    username: string,
+    failures: ValidationFailure[],
+  ) {
+    if (username.length < 8) {
```
jjalan, line 24: We should ensure username is a valid email

src/modules/account/internal/account-param-validation.ts
```diff
@@ -0,0 +1,55 @@
...
+        message: 'Username must be at least 8 characters.',
+      });
+    }
+  }
+
+  public static isPasswordValid(
+    password: string,
+    failures: ValidationFailure[],
+  ) {
+    if (password.length < 8) {
```
jjalan, line 36: We should not use our own custom validation. Best to use something like https://github.com/dropbox/zxcvbn.

We can use this both on frontend and backend

src/modules/account/internal/account-writer.ts
```diff
@@ -10,6 +11,7 @@ export default class AccountWriter {
   ): Promise<Account> {
     try {
       await AccountReader.checkUsernameNotExists(params);
+      AccountParamValidation.validate(params);
```
jjalan, line 14: We should do this before `await AccountReader.checkUsernameNotExists(params);`

jjalan (changes requested):

jjalan (commented):
Test cases are missing.

PR #17, opened 2022-02-18
Heroku deployment setup.

Sparsh-Kumar:
## Description
Changes required for heroku deployment.

## Database schema changes
NA

NA

### Manual test cases run
NA


.github/workflows/main.yml
```diff
@@ -0,0 +1,25 @@
...
+      - 'release/**'
+
+jobs:
+  create_app:
+    runs-on: ubuntu-latest
+    steps:
+      - name: git-checkout
+        uses: actions/checkout@v2
+
+      - name: heroku-create-app
```
jjalan, line 15: We discussed that we will not create app, rather we will only do deployment to avoid issues around managing env variables whenever app is created.

Can we ensure we are not creating new app, but only do a deployment?

tsconfig.json
```diff
@@ -5,7 +5,7 @@
     "moduleResolution": "node",
     "allowSyntheticDefaultImports": true,
     "allowJs": true,
-    "importHelpers": true,
+    "importHelpers": false,
```
jjalan, line 8: Why is this change needed?

.github/workflows/main.yml
Sparsh-Kumar: @jjalan  sir, If the app does not exist then it will create a new app for first time, but after that If again a new release branch is created it will update the app to deploy that new branch.

tsconfig.json
Sparsh-Kumar: we were talking to avoid installing tslib npm package, actually importHelpers use this package to reduce the size of output, https://webhint.io/docs/user-guide/hints/hint-typescript-config/import-helpers/

By default in typescript this option is set to false.

.github/workflows/main.yml
jjalan: Understood.

Let's remove the app creation and only do deploy. I would like that step to be explicit so it would dev a chance to set env variables explicitly.

We should rename it to deploy-app

tsconfig.json
jjalan: Can you also confirm this is also false in Recco project as well?

Sparsh-Kumar: ok

.github/workflows/main.yml
```diff
@@ -0,0 +1,26 @@
...
+        env:
+          HD_NODE_CONFIG_ENV: "qa"
+        with:
+          heroku_api_key: ${{secrets.HEROKU_API_KEY}}
+          heroku_email: ${{secrets.HEROKU_EMAIL}}
+          heroku_app_name: "node-ts-boilerplate-qa"
+          dontautocreate: true
+
+      - name: heroku-add-to-pipeline
+        run: heroku pipelines:add backend-pipeline --app=node-ts-boilerplate-qa --stage=staging || true
```
akashlakhera, line 26: I think the app will already be in a pipeline, no need to add it again.

src/app.ts
```diff
@@ -41,7 +41,7 @@ export default class App {
     // Error handling
     this.app.use(serverErrorHandler);
 
-    const port = ConfigService.getIntValue('server.port');
+    const port = process.env.PORT || ConfigService.getIntValue('server.port');
```
akashlakhera, line 44: Why is this needed? We added `server.port` in custom config.

akashlakhera (changes requested):

.github/workflows/main.yml
Sparsh-Kumar: @akashlakhera  sir, actually there may be a case in which an app exist with this name on Heroku but is not added to the pipeline, so just to make sure that we add the app with this name to the pipeline I have added this step to add the app to pipeline.

src/app.ts
Sparsh-Kumar: yes @akashlakhera  sir, actually here we are trying to get the integer value of port defined in the environment, But on heroku, the value of port is in the form of string therefore we were always getting value mismatch type error, 
so In order to remove that error I have added this line, if the port is defined in environment then it will take value using this ```process.env.PORT```

.github/workflows/main.yml
akashlakhera: What will be the case? Can you give an example? We'll always manually add the add to a pipeline.

src/app.ts
akashlakhera: Then we can do a `getStringValue`. I think every platform/infra provider (AWS, DO, etc) will provide env vars in string.

.github/workflows/main.yml
Sparsh-Kumar: I added it just to be on safe side, no issues I will remove this add to pipeline step.

tsconfig.json
Sparsh-Kumar: yes @jjalan  sir, I have confirmed it we are not setting this flag ```importHelpers``` in recco project

src/app.ts
Sparsh-Kumar: ok @akashlakhera  sir, I think that would work.

.github/workflows/main.yml
Sparsh-Kumar: done

Sparsh-Kumar:
> Can you add manual tests step on how you went about verifying the change?



> Can you add manual tests step on how you went about verifying the change?

Actually @jjalan  sir, in heroku account I cannot create more than 5 apps in free tier and there are already some apps added into that account, so I cannot add another apps in same heroku accounts, so I have messaged @bbbneo333  sir regarding that either he will remove some apps or create another account for heroku.

src/app.ts
```diff
@@ -41,7 +41,7 @@ export default class App {
     // Error handling
     this.app.use(serverErrorHandler);
 
-    const port = process.env.PORT || ConfigService.getIntValue('server.port');
+    const port = ConfigService.getStringValue('server.port');
```
akashlakhera, line 44: I think we'll also need to change the config values to string in yml files.

Sparsh-Kumar: done @akashlakhera  sir.

akashlakhera (approved):

PR #18, opened 2022-02-21
feat: feat: Adding param validators to account api

vindru:
## Description

Adding param validators to account api

## Database schema changes
NA

## Test cases

- POST /account should throw validation error if username is invalid
- POST /account should throw validation error if password is weak


test/spec/account/account.spec.ts
```diff
@@ -51,4 +51,26 @@ describe.skip('Account Service', () => {
...
+  
+  it('POST /account should throw validation error if username is invalid', async () => {
+    const params = { username: 'user', password: 'password' };
+    const res = await chai
+      .request(app)
+      .post('/accounts')
+      .set('content-type', 'application/json')
+      .send(params);
+
+    expect(res.body.httpStatusCode).to.eq(400);
```
jjalan, line 63: Can you assert the validation result array as well?

test/spec/account/account.spec.ts
```diff
@@ -51,4 +51,26 @@ describe.skip('Account Service', () => {
...
+
+  it('POST /account should throw validation error if password is weak', async () => {
+    const params = { username: 'user@test.dev', password: 'pas' };
+    const res = await chai
+      .request(app)
+      .post('/accounts')
+      .set('content-type', 'application/json')
+      .send(params);
+
+    expect(res.body.httpStatusCode).to.eq(400);
```
jjalan, line 74: Can you assert the validation result array as well?

jjalan (changes requested):

src/modules/account/internal/account-param-validation.ts
```diff
@@ -0,0 +1,46 @@
...
+      });
+    }
+  }
+
+  public static isPasswordValid(
+    password: string,
+    failures: ValidationFailure[],
+  ): void {
+    const passwordStrength = zxcvbn(password);
+    if (passwordStrength.score < 3) {
```
jjalan, line 37: It is not clear what `3` stands for.

1. Can you add some comments in the code and reference to other literature what `3` means?
2. Can you also add comment why `3`?

PR #24, opened 2022-07-06
Add support for E2E testing

rayescmata:
## Description
Cypress e2e integration, testing data managers, e2e testing docker-compose

## Database schema changes
none

- login


src/modules/communication/internals/twilio-service.ts
```diff
@@ -5,29 +5,39 @@ import { SendSMSParams, ThirdPartyServiceError } from '../types';
...
 
 export default class TwilioService {
+  private static mock: boolean;
   private static twilio: Twilio;
 
-  public static initializeService(): void {
+  public static initializeService(mockMode: boolean): void {
+    this.mock = mockMode;
+
+    if (this.mock) return;
```
jjalan, line 14: @bbbneo333 

I would highly discourage this pattern.

This means we are adding code that is not desirable even in production build.

Recommend to mock appropriate third party calls using spyOn or any other framework as part of scenarios that get loaded before a E2E test is run. That also gives more flexibility for the caller to define the response from third party (success or error or anything else) depending on the test they want to write.

cypress.config.ts
```diff
@@ -0,0 +1,39 @@
+import { defineConfig } from 'cypress';
+
+import scenarios from './cypress/scenarios';
+
+export default defineConfig({
+  e2e: {
+    /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
```
jjalan, line 7: Highly encourage to not suppress TS warnings.

src/modules/communication/internals/twilio-service.ts
```diff
@@ -35,3 +39,7 @@ export default class TwilioService {
     }
   }
 }
+
+export default ConfigService.getBoolValue('communication.mock')
```
jjalan, line 43: This is still adding the mock if else in production code.

Is it not possible use sinon in scenario testing to mock the third party api call?

bbbneo333:
Overall looks great @rayescmata, as discussed on call, calling out following points:
- For service mocks
  - Need to have layer which sits between a service and the application. This layer would be responsible for returning the correct instance (mocked / unmocked) for the service.
  - We can ditch per service mock configuration and say if e2e env is enabled, return mock for all the services
- Can update the PR title to ticket title - `Add support for E2E testing`

bbbneo333:
Problems with our docker compose files:
- `test` - Assumes that image is already built within the system
- `e2e` - Tries to build it, thus rebuilding the image within CI process

We need to find a common way to define in our compose file that:
- If image is found, use that
- Otherwise built it

bbbneo333:
We can move mocks to `test/mocks`

bbbneo333:
@rayescmata - For demonstration purposes, can you purposefully fail a test case and see if the e2e check actually fails for that commit?

config/development.yml
```diff
@@ -1,4 +1,22 @@
...
+  services: ['twilio', 'sendgrid']
+
+twilio:
+  verify:
+    accountSid: 'AC.TWILIO_ACCOUNT_SID'
+    authToken: 'TWILIO_AUTH_TOKEN'
+  messaging:
+    messagingServiceSid: 'TWILIO_MESSAGING_SERVICE_SID'
+
+sendgridApiKey: 'SG.SENDGRID_API_KEY'
```
bbbneo333, line 16: @rayescmata  - Any reason why this was added? The app should work without these integrations and thus not requiring us to define any configuration values.

config/development.yml
```diff
@@ -1,4 +1,22 @@
...
 
 mongoDb:
-  uri: "mongodb://localhost:27017//boilerplate-mern-dev"
+  uri: 'mongodb://localhost:27017/boilerplate-mern-dev'
+
+server:
+  port: '8080'
```
bbbneo333, line 22: This is being picked up from `default` already. Adding it here will have no effect. Advise to remove this.

config/development.yml
bbbneo333, line 16: @rayescmata  - Any reason why this was added? The app should work without these integrations and thus not requiring us to define any configuration values.

config/e2e.yml
```diff
@@ -0,0 +1,28 @@
+boolTestKey: true
+numberTestKey: 1
+stringTestKey: 'string'
```
bbbneo333, line 3: This should be removed from e2e as this is testing only config (config/testing.yml)

config/e2e.yml
```diff
@@ -0,0 +1,28 @@
...
+  messaging:
+    messagingServiceSid: 'TWILIO_MESSAGING_SERVICE_SID'
+
+sendgridApiKey: 'SG.SENDGRID_API_KEY'
+
+mongoDb:
+  uri: 'mongodb://localhost:27017/boilerplate-mern-e2e'
+
+server:
+  port: '8080'
```
bbbneo333, line 28: As mentioned above, this can be removed as this will get picked up by default (config/default.yml) and this has no effect here.

config/e2e.yml
```diff
@@ -0,0 +1,28 @@
+boolTestKey: true
+numberTestKey: 1
+stringTestKey: 'string'
+
+rollbar:
+  accessToken: 'ROLLBAR_ACCESS_TOKEN'
+  env: 'ROLLBAR_ENV'
```
bbbneo333, line 7: This should be removed as we are not using rollbar in e2e

config/e2e.yml
```diff
@@ -0,0 +1,28 @@
...
+  services: ['twilio', 'sendgrid']
+
+twilio:
+  verify:
+    accountSid: 'AC.TWILIO_ACCOUNT_SID'
+    authToken: 'TWILIO_AUTH_TOKEN'
+  messaging:
+    messagingServiceSid: 'TWILIO_MESSAGING_SERVICE_SID'
+
+sendgridApiKey: 'SG.SENDGRID_API_KEY'
```
bbbneo333, line 22: This should be removed as we are not using these integrations in e2e and thus, not requiring us to define these configuration values.

cypress/e2e/app.spec.cy.ts
```diff
@@ -0,0 +1,44 @@
+let credentials;
```
bbbneo333, line 1: Should be scoped to the `describe` suite

cypress/e2e/app.spec.cy.ts
```diff
@@ -0,0 +1,44 @@
...
+
+describe('application serving', () => {
+  before(() => {
+    // SETUP THE LOGIN SCENARIO
+    cy.task('scenario:setup', 'login').then((creds) => {
+      credentials = creds;
+    });
+    cy.visit('localhost:8080');
```
bbbneo333, line 9: I think there's a way you can define a base url in cypress, and this can be replaced with simple `cy.visit('/')`

cypress/e2e/app.spec.cy.ts
```diff
@@ -0,0 +1,44 @@
+let credentials;
+
+describe('application serving', () => {
+  before(() => {
```
bbbneo333, line 4: Should be `beforeEach` and cleanup should take place in `afterEach`. Even though this can be intensive, but it makes sure every test has clean slate to work on.

cypress/e2e/app.spec.cy.ts
```diff
@@ -0,0 +1,44 @@
...
+    cy.get('button').click();
+
+    cy.get('#error').should('be.visible').should('have.text', 'ERROR!');
+  });
+
+  after(() => {
+    // CLEAN UP THE LOGIN SCENARIO
+    credentials = null;
+    cy.task('scenario:cleanup', 'login');
+  });
```
bbbneo333, line 43: - As mentioned above, use `afterEach`
- As `credentials` will scoped to `describe`, `credentials = null` should not be required

cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "Using fixtures to represent data",
+  "email": "hello@cypress.io",
+  "body": "Fixtures are a great way to mock data for responses to routes"
+}
```
bbbneo333, line 5: If this file is not required, we should remove it.
But as this directory should actually exist, we can have a [.keep](https://stackoverflow.com/questions/29183372/rails-why-is-there-keep-file-in-every-directory) file here.

cypress/mocks/sendgrid-service.ts
```diff
@@ -0,0 +1,13 @@
+import { SendEmailParams } from '../../src/modules/communication/types';
+import EmailParams from '../../src/modules/communication/internals/sendgrid-email-params';
+
+export default class SendGridService {
+  public static initializeService(): void {
+    console.log('SENDGRID MOCK INITIALIZED');
```
bbbneo333, line 6: This should be removed

cypress/support/commands.ts
```diff
@@ -0,0 +1,38 @@
+/* eslint-disable max-len */
```
bbbneo333, line 1: Can this file be removed?

src/server.ts
```diff
@@ -1,9 +1,19 @@
...
   App.startServer().then(
-    () => { },
-    // eslint-disable-next-line no-console
-    (e) => { console.error(e); },
+    (server) => {
+      // eslint-disable-next-line no-console
+      console.log(
+        'Server is listening on port: ',
+        (server.address() as AddressInfo).port,
+      );
```
bbbneo333, line 12: We should not be using `console.log` at any place. We have a common module in place which takes care of logging.

rayescmata:
[https://github.com/jalantechnologies/boilerplate-mern/pull/24#issuecomment-1181419695] @bbbneo333 Here is a pic of the login e2e test that was failed purposefully by changing the values that were input.

Screen Shot 2022-07-12 at 9 29 55 PM

config/development.yml
rayescmata: Removed this already

config/development.yml
rayescmata: Removed this as well

config/e2e.yml
rayescmata: This was left over from the test.yml file but I've removed it

config/e2e.yml
rayescmata: Removed it from the config file

config/e2e.yml
rayescmata: Removed these already

cypress/e2e/app.spec.cy.ts
rayescmata: Added that to the cypress.config.json file

cypress/e2e/app.spec.cy.ts
rayescmata: Made the changes already, but I'm thinking that scenarios should be separate from the general tests.

src/server.ts
rayescmata: Removed this

bbbneo333:
@rayescmata  - Looks good! Let's get it integrated in SE portal and see if our implementation works end-to-end. If any changes are required, will apply them there and get a sign off from team. Will get those changes added here (if applicable) and finally merge this PR.

bbbneo333 (approved):

PR #29, opened 2022-07-20
PT-45: Add support for videos / screenshots for failing test cases

rayescmata:
## Description

Added in support for Cypress Screenshots and Videos. Videos will be recorded each run as Cypress has no configuration to limit this to failing runs. But the screenshots are done on failure.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,7 @@ terraform/terraform.tfstate
 terraform/.terraform.tfstate.lock.info
 terraform/terraform.tfstate.backup
 terraform/*.tfvars
+
+#cypress
+cypress/videos
+cypress/screenshots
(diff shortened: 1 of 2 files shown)


bbbneo333:
@rayescmata  - The changes we have made here, it might help developers to debug the failed test cases when ran locally, but what we more interested here is how we can debug the test cases right away which are failed within the CI process?

Is there a way for us to view the screenshots / videos for failed test cases on here, GitHub actions?

rayescmata:
> Is there a way for us to view the screenshots / videos for failed test cases on here, GitHub actions?

@bbbneo333 I looked it up and unfortunately there doesn't seem to be a readily-available option to upload files (the screenshots and videos) through the Github API or through comment-creating github actions.

But we do have a work-around as we can post URLs like we did with the deployments - So I'm thinking that we could zip the screenshots and videos into a file, upload that file to cloud storage, and just post the url for that file on the PR for download and viewing. What do you think?

PR #36, opened 2022-09-23
PT-28 Use webpack instead of gulp to compile project on boilerplate

akashlakhera:
## Description

Use webpack instead of gulp to compile boilerplate.

## Screenshots

https://user-images.githubusercontent.com/17554101/192758444-a393d12b-240e-4a66-b78f-7b68edd87524.mov


bbbneo333:
@akashlakhera - Encountered following issue when ran `docker compose -f docker-compose.dev.yml up --build`, can you please check?

Screenshot 2022-09-29 at 12 04 42 PM

bbbneo333:
Hey @akashlakhera, while testing out the new build system, noticed that when there's is change in backend, it tries to build the frontend as well (please see video, atleast according to the logs it looks like it but might be wrong). 

https://user-images.githubusercontent.com/23015711/193202100-dc8e370f-94a9-4a09-ab69-bbf5e57f52be.mov

config/webpack/webpack.prod.js
```diff
@@ -0,0 +1,11 @@
+/* eslint-disable import/no-extraneous-dependencies */
```
bbbneo333, line 1: See note from above

config/webpack/webpack.prod.js
```diff
@@ -0,0 +1,11 @@
...
+// eslint-disable-next-line no-unused-vars
+const path = require('path');
```
bbbneo333, line 3: Think these can be removed

bbbneo333 (changes requested):

config/webpack/webpack.common.js
```diff
@@ -0,0 +1,56 @@
...
+      },
+      {
+        test: /\.ts(x)?$/,
+        loader: 'ts-loader',
+        exclude: /node_modules/,
+      },
+    ],
+  },
+  plugins: [
+    new webpack.DefinePlugin({ CONFIG: JSON.stringify(CONFIG) }),
```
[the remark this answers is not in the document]
akashlakhera: This should not inject entire configuration according to [docs](https://github.com/node-config/node-config/wiki/Webpack-Usage#option-1).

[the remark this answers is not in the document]
bbbneo333: Still, don't think this is still something we should be doing. As mentioned:
- Define a `public` config which is dedicated to frontend. Nothing else should injected (irrespective of env)
- The type we have defined for config (`const CONFIG: { [key: string]: { [key: string]: string } };`) should be explicit. 

Say if we have:
```json
{
  "public": {
    "some_config": "some value"
  }
}
```

We should have type definition as:
```typescript
type Config = {
  'some_config': string
};
```

And we should be able to use it as `Config.some_config` in frontend.

src/public/js/ui/app.tsx
```diff
@@ -6,6 +6,7 @@ import InspectLet from '../vendor/inspectlet';
 
 export const App: React.FC = () => {
   useEffect(() => {
+    console.log(CONFIG.server.port);
```
bbbneo333, line 9: Can show the current env on homepage instead

PR #43, opened 2023-01-20
added signup page

Abhiram716:
## Description
created a Signup page which is integrated wit Baseweb.

## UI Screenshot
Screenshot from 2023-01-20 10-04-11


src/apps/frontend/app.component.tsx
```diff
@@ -17,16 +17,19 @@ export default function App(): React.ReactElement {
   }, []);
 
   return (
-    <DepsProvider deps={{
-      accessService: new AccessService(),
-    }}>
+    <DepsProvider
```
jjalan, line 20: Can we get rid of `DepsProvider` and create `AuthContextProvider` etc. Earlier we model `DepsProvider` as one central repository for dependencies but that's not react thinking.

Kindly use reference from Sunlead on this is done and refactor it please.

src/apps/frontend/pages/signup/signup.page.tsx
```diff
@@ -0,0 +1,113 @@
...
+                  autoComplete="off"
+                  name="password"
+                  type="password"
+                  id="password"
+                  placeholder="password"
+                  onChange={formik.handleChange}
+                  value={formik.values.password}
+                  required
+                  overrides={{
+                    Root: {
```
jjalan, line 92: As you will see that you are repeating the code to override here.

It is best to create your own `Input` thats a wrapper around baseeb `Input`.

jjalan (commented):
I would like to see the video with following states as well:

1. Loading: i.e. when the user submits the form, the form should be disabled and it should show a loading indicator
2. Error: When the service returns error or user forgot to enter required information, form should show the right error.

PR #58, opened 2023-08-04
(bug) remove dependency from 'ci_version' input 

Patel-world:
## Description
Testing (bug) remove dependency from 'ci_version' input #18

## Database schema changes
NA
- NA

### Manual test cases run
- _NA


Patel-world:
image
image
I have commented the code that comments here `code coverage` and that is not happened due to commented code meaning its using correct ci version

PR #79, opened 2023-09-06
(feat) Middleware to validate input params before hitting controllers

vjnvisakh-jtc:
## Description
This PR introduces support to validate input params before they reach the controllers via middleware. 
Refer for all possible validations here: https://www.npmjs.com/package/class-validator

## Database schema changes
NA

Omitted purposefully so that teams using this boiler plate are not affected by it. 

### Manual test cases run

`No params passed`
image

`No password passed`
image

`username is not a string`
image


bbbneo333:
This PR is being closed due to inactivity. Our team has logged the issue and this will get taken care of in future released. Thank you for taking out time to raise the PR.

PR #81, opened 2023-09-21
chore: Project updates

bbbneo333:
## Description
This PullRequest adds various enhancements, bug fixes to the boilerplate which are all listed below:

**Application**
- [chore] This introduces following new application level basic abstractions - `applicationController`, `ApplicationRepository`, `ApplicationRouter`, `ApplicationServer`, `ApplicationError`
- [chore] This removes requirements for service managers and bootstrapping / initialising modules. The modules and repositories now can be used without any initialisation. 
- [feat] This updates `Logger` module and adds support for passing additional positional arguments to logger transports.

**Build**
- [chore] This upgrades the node version to `v18` (Hydrogen) and locks it up to `v18.17.1`.

**Config**
- [feat] This adds support for injecting env variables using `.env` file when running the application using `npm start` / `npm run serve`.

**E2E / Cypress**
- [feat] This enables support for capturing screenshots for failed test cases which can be found under `cypress/screenshots`
- [chore] This adds a helper method `setupScenario` to simplify the scenario usage in E2E tests.
- [chore] This adds a new npm script `cy:open` to open up cypress.

**Lint / Eslint**
- [feat] This adds support for alphabetical member ordering via `@typescript-eslint/member-ordering` (TS) and `member-ordering` (JS)
- [chore] This upgrades the `@typescript-eslint/eslint-plugin` and `@typescript-eslint/parser` to latest version - 6.20.0
- [chore] This upgrades the `eslint-config-airbnb-typescript` to latest version - 17.1.0
- [chore] This enables and no longer ignores files from `test` from when running lint checks using eslint.

**Test / Mocha**
- [feat] This adds support for having common logic for registering plugins with `mocha`
- [feat] This adds support for verifying database connection (before running tests) and cleaning up database (before running any test case).
- [feat] This adds support for resetting sinon mocks and stubs after running any test case.
- [chore] This upgrades mocha to latest version - `10.1.0`

**Miscellaneous**
- [chore] This upgrades `jsonwebtoken`, `webpack` to mitigate CVE vulnerability.

## Database schema changes
- NA

- NA

### Manual test cases run
- NA


src/apps/backend/modules/access-token/rest-api/access-token-controller.ts
```diff
@@ -1,29 +1,18 @@
...
-    try {
-      const { username, password }: CreateAccessTokenParams = req.body as CreateAccessTokenParams;
-      const params: CreateAccessTokenParams = { username, password };
-      const accessToken = await AccessTokenService.createAccessToken(params);
-      res.send(AccessTokenController.serializeAccessTokenAsJSON(accessToken));
-    } catch (e) {
-      next(e);
-    }
-  }
+export const createAccessToken = applicationController(async (
```
jjalan, line 7: This seems a bit counter intuitive while reading.

Can you elaborate a bit more on the thinking behind it?

Also the code within `applicationController` is bit involved and not easy to reason with IMO.

src/apps/backend/modules/access-token/rest-api/access-token-serializer.ts
```diff
@@ -0,0 +1,7 @@
+import { AccessToken } from '../types';
+
+export const serializeAccessToken = (accessToken: AccessToken): unknown => ({
```
jjalan, line 3: Recommend to add `AsJSON` here and other places as suffix.

In an ideal world, based on request params (json, csv, xml, etc), we may have different serializers. Adding suffix makes it bit more clear and future proof.

src/apps/backend/modules/access-token/rest-api/account-auth-middleware.ts
```diff
@@ -1,46 +1,41 @@
...
   UnAuthorizedAccessError,
 } from '../types';
 
-export default class AccountAuthMiddleware {
-  public static ensureAccess(
-    req: Request,
-    _res: Response,
-    next: NextFunction,
-  ): void {
+const AccountAuthMiddleware = {
```
jjalan, line 16: should this be `accountAuthMiddleware`, think it better aligns with JS naming convention.

src/apps/backend/modules/logger/internals/grafana-loki-logger.ts
```diff
@@ -1,18 +1,18 @@
 import { format } from 'winston';
 import LokiTransport from 'winston-loki';
 
-import ConfigService from '../../config/config-service';
+import { ConfigService } from '../../config';
 
 import WinstonLogger from './winston-logger';
 
 export default class GrafanaLokiLogger extends WinstonLogger {
```
jjalan, line 8: We should consider removing all types except Datadog which I believe is the official stand moving forward.

bbbneo333: Addressed

src/apps/backend/modules/access-token/rest-api/account-auth-middleware.ts
bbbneo333: Addressed

src/apps/backend/modules/access-token/rest-api/access-token-serializer.ts
bbbneo333: Addressed

src/apps/backend/modules/access-token/types.ts
```diff
@@ -1,24 +1,42 @@
...
+import { HttpStatusCodes } from '../http';
 
 export class AccessToken {
-  accountId: string;
-
-  expiresAt?: Date;
-
   token: string;
+  accountId: string;
+  expiresAt: Date;
```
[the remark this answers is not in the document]
bbbneo333: Addressed

src/apps/backend/modules/access-token/internal/access-token-writer.ts
```diff
@@ -1,39 +1,70 @@
...
-    ) as jsonwebtoken.JwtPayload;
-
-    accessToken.expiresAt = new Date(vetifiedToken.exp * 1000);
+    const jwtTokenDecoded = jsonwebtoken.decode(jwtToken) as jsonwebtoken.JwtPayload;
+    accessToken.expiresAt = new Date(jwtTokenDecoded.exp * 1000);
 
     return accessToken;
   }
+
+  public static verifyAccessToken(
```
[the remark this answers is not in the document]
bbbneo333: Addressed

src/apps/backend/modules/access-token/rest-api/access-token-controller.ts
bbbneo333: Yeah agreed - The current definitions of controllers does not _looks_ right.

On what and why is `applicationController` - This mainly takes care of async controllers, so controllers don't have to do `try/catch`. They can simply define their async functions in there and `applicationController` will take care of handling promise and reporting error if any.

Point can be made to make the name more explicit, according to usage, something like `asyncController`. But was thinking what if we add some defacto logic for controllers in there, something like reporting validation errors before running the actual controller.

This will have all the controller's common logic that every controller can use, hence the name - `applicationController`.

src/apps/backend/modules/application/application-repository.ts
```diff
@@ -0,0 +1,34 @@
...
+  name: string,
+  schema: mongoose.Schema,
+  collection?: string,
+): mongoose.Model<T> {
+  const connection = getDatabaseConnection();
+
+  // TODO: Typings for repositories is not working as expected, this needs
+  //  to be fixed.
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+  return connection.model<T>(name, schema, collection);
```
bbbneo333, line 33: https://github.com/orgs/jalantechnologies/projects/1/views/1?pane=issue&itemId=51722097

src/apps/backend/modules/access-token/rest-api/access-token-controller.ts
jjalan: Would something like `requestHandler` be more appropriate?

jjalan (approved):

config/custom-environment-variables.yml
```diff
@@ -1,22 +1,19 @@
...
-    env: 'NODE_CONFIG_ENV'
+accounts:
+  tokenKey: 'JWT_TOKEN'
 
 rollbar:
   env: 'ROLLBAR_ENV'
   accessToken: 'ROLLBAR_ACCESS_TOKEN'
 
-jwt:
-  token: 'JWT_TOKEN'
```
vjnvisakh-jtc, line 10: `secret` was a better name imo. Also could we add refresh tokens too into the boiler?

config/custom-environment-variables.yml
```diff
@@ -1,22 +1,19 @@
-node:
-  config:
-    env: 'NODE_CONFIG_ENV'
+accounts:
+  tokenKey: 'JWT_TOKEN'
```
vjnvisakh-jtc, line 2: `secret` under jwt was better imo. Also could we introduce refresh tokens as well into the boiler?

config/default.yml
```diff
@@ -1,11 +1,15 @@
-jwt:
-  expiresIn: '1y' # default token expire time 1 year. Examples - https://github.com/vercel/ms#examples
+accounts:
+  tokenExpiry: '1y' # default token expire time 1 year. Examples - https://github.com/vercel/ms#examples
```
vjnvisakh-jtc, line 2: imo, access token expiries should be short say a day etc. This way even if it falls in the wrong hands it can't do much damage.

src/apps/backend/modules/access-token/access-token-service.ts
```diff
@@ -1,10 +1,69 @@
...
+        jwtSigningKey,
+        { ignoreExpiration: true },
+      ) as jsonwebtoken.JwtPayload;
+    } catch (e) {
+      throw new AccessTokenInvalidError();
+    }
+
+    if (verifiedToken.exp * 1000 < Date.now()) {
+      throw new AccessTokenExpiredError();
+    }
```
vjnvisakh-jtc, line 63: The .verify also throws an error if the token has expired

src/apps/backend/modules/access-token/index.ts
vjnvisakh-jtc, line 1: just a suggestion. `auth` or `authentication` module would have been a good name

src/apps/backend/modules/access-token/rest-api/access-token-router.ts
```diff
@@ -1,14 +1,11 @@
...
+import * as AccessTokenController from './access-token-controller';
 
-export default class AccessTokenRouter {
-  public static getRoutes(): Router {
-    const router = Router();
+export default class AccessTokenRouter extends ApplicationRouter {
+  configure(): void {
+    const { router } = this;
 
     router.post('/', AccessTokenController.createAccessToken);
```
vjnvisakh-jtc, line 9: just for some added safety, would it be possible to configure a list of domains from which we allow access? 

Seen this being done - 
image

src/apps/backend/modules/account/internal/store/account-repository.ts
```diff
@@ -1,30 +1,10 @@
-import mongoose, { CallbackError, Connection } from 'mongoose';
+import { ApplicationRepository } from '../../../application';
 
-import ConfigService from '../../../config/config-service';
+import { AccountDB, AccountDbSchema } from './account-db';
 
-import { AccountDB, accountDbSchema } from './account-db';
+const AccountRepository = ApplicationRepository<AccountDB>(
```
vjnvisakh-jtc, line 5: nice  😊

src/apps/backend/modules/communication/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default as EmailService } from './email-service';
+export { default as SMSService } from './sms-service';
```
vjnvisakh-jtc, line 2: could we add a push notification service as well by default?

src/apps/backend/modules/task/rest-api/task-router.ts
```diff
@@ -1,18 +1,17 @@
...
-    router.post('/', AccountAuthMiddleware.ensureAccess, TaskController.createTask);
-    router.get('/', AccountAuthMiddleware.ensureAccess, TaskController.getAllTasks);
-    router.get('/:id', AccountAuthMiddleware.ensureAccess, TaskController.getTask);
-    router.delete('/:id', AccountAuthMiddleware.ensureAccess, TaskController.deleteTask);
-
-    return router;
+    router.post('/', TaskController.createTask);
+    router.get('/', TaskController.getTasks);
+    router.get('/:id', TaskController.getTask);
+    router.delete('/:id', TaskController.deleteTask);
```
vjnvisakh-jtc, line 15: I have noticed at times that if you give `/abc/:abcid` as a common prefix in the `*-server.ts` file and then append something in the `*-router.ts` file it doesn't quite work.

test/hooks/database.ts
```diff
@@ -0,0 +1,31 @@
+import config from 'config';
+import mongoose from 'mongoose';
+
+const mongoDbURI = config.get('mongoDb.uri');
+
+export const cleanDatabase = async (): Promise<void> => {
```
vjnvisakh-jtc, line 6: nice 😊

vjnvisakh-jtc (approved):

config/custom-environment-variables.yml
bbbneo333: > secret under jwt was better imo
@vjnvisakh-jtc - By `secret`, do you mean `jwt.token` ? If yes - `jwt.token` configuration is too generic and I have seen developers using this config for every JWT that they sign, which is not a right thing to do. Any specific reason why was it better?

> Also could we introduce refresh tokens as well into the boiler?
That is a planned item and it out of the scope of this PR.

src/apps/backend/modules/task/rest-api/task-router.ts
bbbneo333: Can you provide some more details on this? I don't have much detail here on the issue in order to address it here.

src/apps/backend/modules/communication/index.ts
bbbneo333: You are encouraged to raise an issue on this repo. This would be out of scope of this PR.

src/apps/backend/modules/access-token/access-token-service.ts
bbbneo333: Yes it does, but we still need to massage to one of our application errors in order to propagate it further up.

src/apps/backend/modules/communication/index.ts
vjnvisakh-jtc: ok

src/apps/backend/modules/access-token/rest-api/access-token-router.ts
vjnvisakh-jtc: was pointing this way only our frontend clients (web and mobile) can access the backend apis

src/apps/backend/modules/access-token/access-token-service.ts
vjnvisakh-jtc: ok. verify throws error with a specific code. the if block can be moved there too. But yeah, `if` block is needed there too.

config/custom-environment-variables.yml
vjnvisakh-jtc: I always thought `secret` as something used for signing and `token` as the end result. Okay with me.

jjalan (approved):

config/default.yml
bbbneo333: Addressed - Changed expiry to `1d`

src/apps/backend/modules/access-token/index.ts
bbbneo333: Good suggestion, I will be raising a followup PR which would probably include this.

PR #92, opened 2024-01-11
Removes bootstrap and add support for Baseweb ui

balwant-jtc:
## Description
Removes bootstrap and add support for Baseweb ui

https://github.com/jalantechnologies/boilerplate-mern/assets/78337443/15897f82-6e1b-45dc-9a14-809a2d531d8d


src/apps/frontend/app.component.tsx
```diff
@@ -20,20 +19,22 @@ export default function App(): React.ReactElement {
...
-            <Route path='/about' element={<About />} />
-            <Route path='/' element={<Login />} />
-            <Route path='*' element={<NotFound />} />
-          </Routes>
-          <Footer />
-        </div>
-      </Router>
-    </DepsProvider>
+    <BaseWebProvider>
+      <DepsProvider
```
jjalan, line 23: We should deprecate `DepsProvider`. We initially thought to use dependency injection and `DepsProvider` knows all about application dependencies, kinda inspired by angular. React has its own context paradigm which is better suited, and hence this should be deprecated

src/apps/frontend/app.global.css
```diff
@@ -0,0 +1,29 @@
+/* add your global imports and css rules her */
```
jjalan, line 1: Where is this coming from?

I believe with baseweb, there should not be any need to write css like this, except when we work with 3rd part lib and we need to override some of the css directly (which again should be discouraged).

src/apps/frontend/hooks/login-form.hook.ts
```diff
@@ -0,0 +1,50 @@
...
+  const [success, setSuccess] = useState(false);
+  const [error, setError] = useState(false);
+
+  const login = useCallback(
+    async ({ username, password }) => {
+      setSuccess(false);
+      setError(false);
+
+      try {
+        const accessService = new AccessService();
```
jjalan, line 18: You can move the initialization of this outside this UI component i.e. 

```
const accessService = new AccessService();
const useLoginForm = () => {
```

Should we also move this closer to the login form, and avoid creating another folder `hooks`? I can see in future everyone would put all context here which would make this folder explode

also, `AccessService` seems mislead, recommend to rename to better reflect what it is.

src/apps/frontend/hooks/login-form.hook.ts
```diff
@@ -0,0 +1,50 @@
...
+        await accessService.login(username, password);
+        setSuccess(true);
```
jjalan, line 20: You want to consider storing the token and response into local storage, so that once user login, it can be read from it. You want to expose logout button to then clear local storage.

This can also be considered as a separate PR in case there is no logged in page today that exists after user login.

src/apps/frontend/pages/authentication/login/login-form.tsx
```diff
@@ -0,0 +1,115 @@
...
+          </Checkbox>
+        </Block>
+        <Block>
+          <Button size={SIZE.default} overrides={{
+            BaseButton: {
+              style: () => ({
+                width: '100%',
+              }),
+            },
+          }}
```
jjalan, line 125: hmm.

I assume there is no property on basweb that expands button to take full width? Does setting display type as block on the button also does not achieve that?

src/apps/frontend/pages/authentication/login/login-form.tsx
```diff
@@ -0,0 +1,115 @@
...
+            onChange={formik.handleChange}
+            labelPlacement={LABEL_PLACEMENT.right}
+            overrides={{
+              Checkmark: {
+                style: ({ $theme }) => ({
+                  backgroundColor: `${$theme.colors.inputBackground}`,
+                  border: `1px solid ${$theme.border.borderColor}`,
+                  borderRadius: '0.25em',
+                }),
+              },
```
jjalan, line 112: Override globally or create custom component with the overrides. A page should not create overrides for primitive components

src/apps/frontend/pages/authentication/login/login-form.tsx
```diff
@@ -0,0 +1,115 @@
...
+              placeholder="Password"
+              value={formik.values.password}
+              onChange={formik.handleChange}
+              onBlur={formik.handleBlur}
+              autoComplete="off"
+              type="password"
+            />
+          </FormControl>
+        </Block>
+        <Block style={{ marginTop: '1rem', marginBottom: '1rem' }}>
```
jjalan, line 99: Use theme spacing class to control margins and paddings.

Also I believe you want to set rowGap on the Block that contains the input form to create consistent spacing between each bloc element.

src/apps/frontend/pages/authentication/login/login-form.tsx
```diff
@@ -0,0 +1,115 @@
...
+              autoComplete="off"
+            />
+          </FormControl>
+        </Block>
+        <Block>
+          <FormControl label={() => 'Password'} error={formik.errors.password}>
+            <Input
+              overrides={{
+                Root: {
+                  props: {
```
jjalan, line 77: Remove please

src/apps/frontend/pages/authentication/login/login-form.tsx
```diff
@@ -0,0 +1,115 @@
...
+                    'data-testid': 'password',
+                  },
+                },
+                Input: {
```
jjalan, line 81: Override globally or create custom component with the overrides. A page should not create overrides for primitive components

src/apps/frontend/pages/authentication/login/login-form.tsx
```diff
@@ -0,0 +1,115 @@
...
+        <Block>
+          <FormControl
+            label={() => 'Email address'}
+            error={formik.errors.username}
+          >
+            <Input
+              overrides={{
+                Root: {
+                  props: {
+                    'data-testid': 'username',
```
jjalan, line 42: Remove

src/apps/frontend/pages/authentication/login/login-form.tsx
```diff
@@ -0,0 +1,115 @@
...
+                  },
+                },
+                Input: {
+                  style: ({ $theme }) => ({
```
jjalan, line 46: Override globally or create custom component with the overrides. A page should not create overrides for primitive components

src/apps/frontend/hooks/login-form.hook.ts
```diff
@@ -0,0 +1,50 @@
+import { useFormik } from 'formik';
+import { useState, useCallback } from 'react';
+import * as Yup from 'yup';
+
+import constant from '../constants';
+import { AccessService } from '../services';
+
+const useLoginForm = () => {
+  const [success, setSuccess] = useState(false);
+  const [error, setError] = useState(false);
```
jjalan, line 10: Recommend creating [this](https://teams.microsoft.com/l/message/19:c388dc298f0f432bb4805cadc4ffe92f@thread.tacv2/1705932812553?tenantId=79836b2a-53cc-4854-81b4-ba2d7c9f2726&groupId=c0e79bd0-36ec-4b10-8be7-55f5b3d2395c&parentMessageId=1705932812553&teamName=Jalan%20Technologies&channelName=Engineering&createdTime=1705932812553) abstraction

src/apps/frontend/pages/authentication/login/login-form.tsx
Balwant-Shekhawat: No, it didn't worked

src/apps/frontend/hooks/index.ts
```diff
@@ -0,0 +1,4 @@
+/* eslint-disable import/prefer-default-export */
```
jjalan, line 1: Can we either consider updating lint rules or follow the guidelines so we don't have to suppress here and other places?

src/apps/frontend/pages/authentication/login/login-form.tsx
```diff
@@ -0,0 +1,137 @@
...
+      <Block>
+        <Button
+          size={SIZE.default}
+          overrides={{
+            BaseButton: {
+              style: () => ({
+                width: '100%',
+              }),
+            },
+          }}
```
jjalan, line 129: For now, let's remove this and center the button.

src/apps/frontend/pages/authentication/login/login-form.tsx
```diff
@@ -0,0 +1,137 @@
...
+                border: `1px solid ${$theme.border.borderColor}`,
+                borderRadius: '0.25em',
+              }),
+            },
+            Label: {
+              style: ({ $theme }) => ({
+                color: `${$theme.colors.fontColor}`,
+              }),
+            },
+          }}
```
jjalan, line 115: No instance based overrides please

src/apps/frontend/hooks/login-form.hook.ts
```diff
@@ -0,0 +1,50 @@
...
+      username: '',
+      password: '',
+      rememberMe: false,
+    },
+    validationSchema: Yup.object({
+      username: Yup.string()
+        .email(constant.EMAIL_VALIDATION_ERROR)
+        .required('Required'),
+      password: Yup.string()
+        .min(constant.PASSWORD_MIN_LENGTH, constant.PASSWORD_VALIDATION_ERROR)
```
jjalan, line 39: Please use this for password validation - https://github.com/dropbox/zxcvbn

src/apps/frontend/hooks/index.ts
Balwant-Shekhawat: I choose to suppress the warning here because in the future, we might want to export multiple named items from the index.ts file. Currently, we're only exporting a single hook, and TypeScript is reminding us to use the default export. However, if we comply with this warning now, it could lead to unnecessary changes throughout the codebase when we decide to export multiple items later on.

And if we chose to suppress the warning from the TypeScript configuration. This way, we won't be alerted even if we later export just a single item from a file. Which is also not correct.

While generally not a fan of suppressing warnings, in this specific situation, it seems reasonable to do so for the sake of future flexibility and to prevent unnecessary code modifications.

 Interestingly, Baseweb's codebase also includes instances where these TypeScript warnings are suppressed.

src/apps/frontend/pages/authentication/login/login-form.tsx
Balwant-Shekhawat: Used a boolean prop to handle this

src/apps/frontend/pages/authentication/login/login-form.tsx
Balwant-Shekhawat: Added just to align with the style of the page, if you want i can remove the override part but then it will look different from the page.

src/apps/frontend/hooks/login-form.hook.ts
Balwant-Shekhawat: Will try to come up with it, but for now resuming on Valon's tasks.

src/apps/frontend/hooks/login-form.hook.ts
Balwant-Shekhawat: Done, renamed it to AuthService

src/apps/frontend/hooks/login-form.hook.ts
Balwant-Shekhawat: Will do this after implementing [this](https://teams.microsoft.com/l/message/19:c388dc298f0f432bb4805cadc4ffe92f@thread.tacv2/1705932812553?tenantId=79836b2a-53cc-4854-81b4-ba2d7c9f2726&groupId=c0e79bd0-36ec-4b10-8be7-55f5b3d2395c&parentMessageId=1705932812553&teamName=Jalan%20Technologies&channelName=Engineering&createdTime=1705932812553)

src/apps/frontend/pages/authentication/login/login-form.tsx
Balwant-Shekhawat: But now handled it with the boolean prop `fullWidth`

src/apps/frontend/pages/authentication/login/login-form.tsx
Balwant-Shekhawat: Created Custom Component

src/apps/frontend/pages/authentication/login/login-form.tsx
Balwant-Shekhawat: tried using `GridRowGap` but it didn't worked, so used theme spacing to make it work.

src/apps/frontend/pages/authentication/login/login-form.tsx
Balwant-Shekhawat: Done

src/apps/frontend/pages/authentication/login/login-form.tsx
Balwant-Shekhawat: Created custom component

src/apps/frontend/pages/authentication/login/login-form.tsx
Balwant-Shekhawat: Done

src/apps/frontend/pages/authentication/login/login-form.tsx
Balwant-Shekhawat: Created Custom Component

src/apps/frontend/types/async-operation.ts
```diff
@@ -0,0 +1,18 @@
...
+  Idle = 'idle',
+  Pending = 'pending',
+  Success = 'success',
+  Error = 'error',
+}
+
+export interface AsyncOperation<T> {
+  result?: T;
+  error?: ServiceError;
+  status: AsyncOperationStatus;
```
jjalan, line 13: Async operation is an upper level abstraction than Service Error. This does not sound right

src/apps/frontend/types/async-operation.ts
```diff
@@ -0,0 +1,18 @@
+import { ServiceError, ServiceResponse } from './service-response';
+
+export enum AsyncOperationStatus {
+  Idle = 'idle',
+  Pending = 'pending',
+  Success = 'success',
+  Error = 'error',
```
jjalan, line 7: Should they be all uppercase?

src/apps/frontend/pages/authentication/login/login-form.tsx
```diff
@@ -0,0 +1,75 @@
...
+import useLoginForm from './login-form.hook';
+
+export default function LoginForm(): React.ReactElement {
+  const {
+    formik, status, error,
+  } = useLoginForm();
+
+  return (
+    <form onSubmit={formik.handleSubmit}>
+      {status !== AsyncOperationStatus.Idle ? (
```
jjalan, line 20: This does not sound right - I think if operation is success, then it should show success

jjalan (commented):
Can you add a demo video of the new login experience?

src/apps/frontend/types/async-operation.ts
```diff
@@ -0,0 +1,18 @@
...
+}
+
+export interface AsyncOperation<T> {
```
jjalan, line 10: `K, AsyncError<Y>`

src/apps/frontend/services/auth.service.ts
```diff
@@ -0,0 +1,21 @@
+import { AccessToken, ServiceError, ServiceResponse } from '../types';
+
+import APIService from './api.service';
+
+export default class AuthService extends APIService {
+  login = async (
+    username: string,
+    password: string,
+  ): Promise<ServiceResponse<AccessToken>> => {
```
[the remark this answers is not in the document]
balwant-jtc: You mean `Promise<AsyncResponse<AccessToken, ApiError>>`? 
I can do this by doing `class ServiceResponse implements AsyncResponse` but then `AsyncResponse` has a `status` property. The service layer won't handle the status; instead, it will be managed by the `useAsync` hook, which sets it to values like `IDLE, PENDING, SUCCESS, or ERROR`."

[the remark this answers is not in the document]
jjalan: > The service layer won't handle the status

Should it?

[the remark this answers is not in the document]
balwant-jtc: IMO No, because even if service change the status from Pending to Sucess or Error, UI won't get updated. Also then `useAsync` would be of no use as it cant be used inside a service.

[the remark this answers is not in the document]
balwant-jtc: The service layer should not have dependencies or concerns related to the UI layer.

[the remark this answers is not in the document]
jjalan: I think `status` is not a concern of UI layer, rather its an inherent property of any async operation.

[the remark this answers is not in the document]
balwant-jtc: But then it should be limited to SUCCESS or ERROR only. And there will be a middle layer(hook) which will expose `loading: boolean` to UI layer

src/apps/frontend/types/async-operation.ts
balwant-jtc: Addressed

src/apps/frontend/types/async-operation.ts
balwant-jtc: Done

src/apps/frontend/pages/authentication/login/login-form.tsx
balwant-jtc: Done

src/apps/frontend/types/async-operation.ts
balwant-jtc: Done

src/apps/frontend/components/checkbox/checkbox.component.tsx
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable @typescript-eslint/restrict-template-expressions */
+import {
+  Checkbox as BaseCheckbox,
+  CheckboxProps as BaseCheckboxProps,
+  LABEL_PLACEMENT,
+} from 'baseui/checkbox';
+import React from 'react';
+
+export interface CheckboxProps extends BaseCheckboxProps {
+  testId: string;
```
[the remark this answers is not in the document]
balwant-jtc: Done

src/apps/frontend/components/button/button.component.tsx
```diff
@@ -0,0 +1,29 @@
+/* eslint-disable @typescript-eslint/restrict-template-expressions */
+import {
+  Button as BaseButton, ButtonProps as BaseButtonProps,
+} from 'baseui/button';
+import React from 'react';
+
+export interface ButtonProps extends BaseButtonProps {
+  testId?: string;
```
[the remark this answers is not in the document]
balwant-jtc: Done

src/apps/frontend/app.global.css
```diff
@@ -0,0 +1,14 @@
+* {
```
jjalan, line 1: We want to avoid this file

src/apps/frontend/components/form-input/form-input.component.tsx
```diff
@@ -0,0 +1,23 @@
...
+const FormControl: React.FC<React.PropsWithChildren<FormControlProps>> = ({
+  children,
+  ...props
+}) => (
+  <BaseFormControl
+    {...props}
+    overrides={{
+      Label: {
+        style: ({ $theme }) => ({
+          color: `${$theme.colors.fontColor}`,
```
jjalan, line 14: Why do we need to override this? Can we not and use the form control directly at consumer?

src/apps/frontend/components/input/input.component.tsx
```diff
@@ -0,0 +1,34 @@
...
+    overrides={{
+      Input: {
+        props: {
+          'data-testid': testId,
+        },
+        style: ({ $theme }) => ({
+          backgroundColor: `${$theme.colors.inputBackground}`,
+          color: `${$theme.colors.fontColor}`,
+          caretColor: `${$theme.colors.fontColor}`,
+        }),
```
jjalan, line 28: Why do we need to override this? Can we not and use the form control directly at consumer?

src/apps/frontend/contexts/async.hook.ts
```diff
@@ -0,0 +1,45 @@
...
+import {
+  AsyncResult,
+  AsyncOperationStatus,
+  ApiResponse,
+  UseAsyncReturn,
+  ApiError,
+} from '../types';
+
+const useAsync = <T>(): UseAsyncReturn<T> => {
+  const [asyncOperation, setAsyncOperation] = useState<
```
jjalan, line 12: rename to `asyncOperationResult`

src/apps/frontend/pages/authentication/login/login-form.hook.ts
```diff
@@ -0,0 +1,49 @@
...
+import constant from '../../../constants';
+import { useAuthContext } from '../../../contexts';
+import useAsync from '../../../contexts/async.hook';
+
+const useLoginForm = () => {
+  const { login: signIn } = useAuthContext();
+  const {
+    status, error, result, apiCallback,
+  } = useAsync();
+
```
jjalan, line 14: This semantic seems a bit odd.

I would expect it to be:

```
const {
    status, error, result, asyncOperation: as login
  } = useAsync(signIn(username, password)));
```

This should not be needed:

```
const login = useCallback(
    ({ username, password }) => {
      apiCallback(() => signIn(username, password));
    },
    [apiCallback, signIn],
  );
```

src/apps/frontend/theme.ts
```diff
@@ -0,0 +1,23 @@
...
+    buttonPrimaryHover: primitives.accent600,
+    backgroundPrimary: primitives.mono300,
+    fontColor: primitives.white,
+    inputBackground: primitives.mono400,
+    inputPlaceholder: primitives.white,
+  },
+  border: {
+    borderColor: primitives.mono100,
+  },
+};
```
jjalan, line 23: For now, can we not create any overrides, and use whatever comes from basweb?

src/apps/frontend/types/async-operation.ts
```diff
@@ -0,0 +1,18 @@
...
+  PENDING = 'PENDING',
+  SUCCESS = 'SUCCESS',
+  ERROR = 'ERROR',
+}
+
+export interface AsyncResult<T, AsyncError> {
+  result?: T;
+  error?: AsyncError;
+  status: AsyncOperationStatus;
+}
```
jjalan, line 14: order alphabetically

jjalan:
1. Can we keep login page background white?
2. As you soon as you enter username, password field also showed error, this is not desireable as user did not even worked with that field.
3. The word `Success` is ugly. Can we use a toast? This is - I am assuming there is no logged in page right now which shows a header with username and ability to logout.

src/apps/frontend/app.global.css
balwant-jtc: The purpose of including this file is to manage the margin and padding of the body, a task that, in my opinion, BaseWeb cannot handle.

jjalan: Possibly yes, but this opens up and can often act as a reference that its ok to include raw css

src/apps/frontend/pages/authentication/login/login-form.hook.ts
Balwant-Shekhawat: refactored it.

src/apps/frontend/contexts/async.hook.ts
Balwant-Shekhawat: Refactored the hook

src/apps/frontend/components/input/input.component.tsx
Balwant-Shekhawat: Done

src/apps/frontend/components/form-input/form-input.component.tsx
Balwant-Shekhawat: Done

src/apps/frontend/app.global.css
Balwant-Shekhawat: Added a note saying `AVOID ADDING MORE CSS HERE. UTILIZE BASEWEB STYLING INSTEAD` in this file but didn't removed this file because i didn't find a way to change margin padding of body through baseweb

src/apps/frontend/types/async-operation.ts
Balwant-Shekhawat: Done

balwant-jtc:
> 1. Can we keep login page background white?
> 2. As you soon as you enter username, password field also showed error, this is not desireable as user did not even worked with that field.
> 3. The word `Success` is ugly. Can we use a toast? This is - I am assuming there is no logged in page right now which shows a header with username and ability to logout.

Addressed all points.

balwant-jtc:
> 1. Can we keep login page background white?
> 2. As you soon as you enter username, password field also showed error, this is not desireable as user did not even worked with that field.
> 3. The word `Success` is ugly. Can we use a toast? This is - I am assuming there is no logged in page right now which shows a header with username and ability to logout.

Addressed all points.

src/apps/frontend/theme.ts
balwant-jtc: Removed the primitives and overrides, and used default styling.

src/apps/frontend/app.global.css
jjalan: What happens if we don't remove it for time being?

src/apps/frontend/pages/authentication/login/login-form.tsx
```diff
@@ -0,0 +1,84 @@
+/* eslint-disable @typescript-eslint/restrict-template-expressions */
```
bbbneo333, line 1: We are disabling this in multiple files. What is about and can we look into addressing it?

bbbneo333 (approved):

src/apps/frontend/components/button/button.component.tsx
```diff
@@ -0,0 +1,27 @@
+/* eslint-disable @typescript-eslint/restrict-template-expressions */
```
jjalan, line 1: Can we please avoid this?

src/apps/frontend/pages/authentication/login/login-form.hook.ts
```diff
@@ -0,0 +1,43 @@
...
+import useAsync from '../../../contexts/async.hook';
+
+const useLoginForm = () => {
+  const { login: signIn } = useAuthContext();
+  const {
+    isLoading,
+    error,
+    result,
+    apiCallback: loginCallback,
+  } = useAsync(signIn);
```
jjalan, line 15: `const { isLoginLoading, loginError, loginResult, login} = useAuthContext();`

src/apps/frontend/pages/authentication/login/login-form.tsx
jjalan, line 1: Please find a way to remove this

src/apps/frontend/pages/authentication/login/login-form.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { HeadingMedium } from 'baseui/typography';
+import React, { useEffect } from 'react';
+
+import {
+  Button,
+} from '../../../components';
+
+import useLoginForm from './login-form.hook';
+
+export default function LoginForm(): React.ReactElement {
```
jjalan, line 15: `export default LoginForm(): React.ReactElement = () => {`

src/apps/frontend/pages/authentication/login/login-form.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const {
+    formik, isLoading, error, result,
```
jjalan, line 17: `form`

Later, it should probably be used as:

`<form {...form}>`

src/apps/frontend/pages/authentication/login/index.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+import AuthenticationFormLayout from '../authentication-form-layout';
+import AuthenticationPageLayout from '../authentication-page-layout';
+
+import LoginForm from './login-form';
+
+export const Login: React.FC = () => (
+  <AuthenticationPageLayout>
+    <AuthenticationFormLayout>
+      <LoginForm></LoginForm>
```
jjalan, line 11: `onSuccess`, `onError`.

The page should be responsible to show error if its an error via toast, or navigate to `/` (or show toast for now) if success

jjalan: `<LoginForm onSuccess ={onSuccess} onError ={onError}></LoginForm>`

src/apps/frontend/types/async-operation.ts
```diff
@@ -0,0 +1,11 @@
+import { ApiError } from './service-response';
+
+export interface AsyncResult<T, AsyncError> {
+  error?: AsyncError;
+  isLoading: boolean;
+  result?: T;
+}
+
+export interface UseAsyncResponse<T> extends AsyncResult<T, ApiError> {
```
jjalan, line 9: This is a high level construct, should not anything about api.

src/apps/frontend/pages/authentication/login/login-form.tsx
```diff
@@ -0,0 +1,84 @@
...
+  } = useLoginForm();
+
+  useEffect(() => {
+    if (!isLoading && error) {
+      toaster.negative(error.message);
+    } else if (!isLoading && result) {
+      toaster.positive('Login Successful');
+    }
+  }, [isLoading, error, result]);
```
jjalan, line 26: We should remove this.

src/apps/frontend/pages/authentication/login/login-form.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  return (
+    <form onSubmit={formik.handleSubmit}>
+      <HeadingMedium marginBottom="scale600">Please sign in</HeadingMedium>
+
+      <Block>
+        <FormControl
+          label={() => 'Email address'}
```
jjalan, line 34: label="Email Address"

src/apps/frontend/pages/authentication/login/login-form.tsx
```diff
@@ -0,0 +1,84 @@
...
+                  'data-testid': 'username',
+                },
+              },
+            }}
+            name="username"
+            placeholder="Email address"
+            value={formik.values.username}
+            onChange={formik.handleChange}
+            onBlur={formik.handleBlur}
+            autoComplete="off"
```
jjalan, line 50: order alphabetically.

balwant-jtc: Done

src/apps/frontend/pages/authentication/login/login-form.tsx
balwant-jtc: Done

src/apps/frontend/pages/authentication/login/login-form.tsx
```diff
@@ -0,0 +1,84 @@
...
+            value={formik.values.password}
+            onChange={formik.handleChange}
+            onBlur={formik.handleBlur}
+            autoComplete="off"
+            type="password"
+          />
+        </FormControl>
+      </Block>
+      <Block>
+        <Button type="submit" fullWidth disabled={isLoading}>
```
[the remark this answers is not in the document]
balwant-jtc: Done, `type` is not a property in baseweb so removed it from form as well.

src/apps/frontend/pages/authentication/login/login-form.tsx
balwant-jtc: Done

src/apps/frontend/types/async-operation.ts
balwant-jtc: Done

src/apps/frontend/pages/authentication/login/login-form.tsx
balwant-jtc: Done

jjalan (commented):
Can you please update the video to show the error scenario as well, ex - say your laptop does not have internet connectivity or say the backend api throws an error?

src/apps/frontend/pages/authentication/login/login-form.tsx
balwant-jtc: Done

src/apps/frontend/pages/authentication/login/login-form.tsx
balwant-jtc: Done

src/apps/frontend/pages/authentication/login/login-form.hook.ts
balwant-jtc: Done

src/apps/frontend/components/button/button.component.tsx
balwant-jtc: Done

src/apps/frontend/contexts/async.hook.ts
```diff
@@ -0,0 +1,55 @@
...
+    },
+    [asyncFn],
+  );
+
+  useEffect(() => {
+    if (immediate) {
+      // eslint-disable-next-line @typescript-eslint/no-floating-promises
+      asyncCallback();
+    }
+  }, [asyncCallback, immediate]);
```
jjalan, line 45: Why is this needed? I see `immediate` as always false

balwant-jtc: If we want the function to run when the component mounts, we can set this flag to true. That way, we don't have to call the asyncCallback ourselves.

balwant-jtc: For instance, if we need to retrieve the profile information of the logged-in user, we can directly use:

`const {isLoading, error, result } = useAuthContext(userProfile)`

There's no requirement to write it like this:

```
const {isLoading, error, result, userProfileCallback } = useAuthContext(userProfile)

userProfileCallback()
```

balwant-jtc: Not using it at the moment.

src/apps/frontend/pages/authentication/login/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const onError = (error) => {
+    toaster.negative(error.message);
+  };
+
+  const { formik, isLoginLoading } = useLoginForm({ onSuccess, onError });
+
+  return (
+    <AuthenticationPageLayout>
+      <AuthenticationFormLayout>
+        <LoginForm isLoginLoading={isLoginLoading} {...formik}></LoginForm>
```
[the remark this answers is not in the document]
balwant-jtc: then how can i achieve this? : https://github.com/jalantechnologies/boilerplate-mern/pull/92#discussion_r1477759442

src/apps/frontend/types/async-operation.ts
```diff
@@ -0,0 +1,39 @@
+import constant from '../constants';
+
+import { JsonObject } from './common-types';
+
+export class AsyncError {
+  code: string;
+  httpStatusCode: number;
```
jjalan, line 7: This is not right. AsyncError cannot have httpStatusCode

src/apps/frontend/contexts/async.hook.ts
```diff
@@ -0,0 +1,50 @@
...
+
+        if (response.data) {
+          setResult({ ...response.data });
+        }
+
+        return response?.data;
+      } catch (e) {
+        const err = e?.response?.data?.httpStatusCode
+          ? new AsyncError(e?.response?.data)
+          : new DefaultAsyncError(e);
```
jjalan, line 32: async hook should not know anything about httpStatusCode

src/apps/frontend/types/async-operation.ts
balwant-jtc: Removed

src/apps/frontend/contexts/async.hook.ts
balwant-jtc: Removed

src/apps/frontend/pages/authentication/login/login-form.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+import {
+  Button,
+} from '../../../components';
+
+import useLoginForm from './login-form.hook';
+
+interface LoginFormProps {
+  onSuccess: () => void;
+  onError: (error: string) => void;
```
jjalan, line 16: Can we pass error object instead of string?

src/apps/frontend/pages/authentication/login/login-form.tsx
```diff
@@ -0,0 +1,85 @@
...
+}
+
+const LoginForm: React.FC<LoginFormProps> = ({ onError, onSuccess }) => {
```
jjalan, line 19: I am not sure, can you confirm - 

When user submitted the form - the form controls are disabled, including login button as button also shows loading icon?

jjalan (approved):

src/apps/frontend/pages/authentication/login/login-form.tsx
balwant-jtc: Form controls are not disabled only login button is disabled with loading icon.

jjalan: We should disable form controls as well. When the login is in process we do not want user to edit their information.

src/apps/frontend/pages/authentication/login/login-form.tsx
balwant-jtc: Done

src/apps/frontend/pages/authentication/login/login-form.tsx
balwant-jtc: Done

PR #99, opened 2024-02-12
Deprecate baseweb and implement tailwind css theme

balwant-jtc:
## Description
As we are moving away from baseweb to Tailwind, all the baseweb components needs to be replaced by TailAdmin's. This PR addresses deprecation of baseweb and implemention of tailwind css([TailAdmin](https://tailadmin.com/))

https://github.com/jalantechnologies/boilerplate-mern/assets/74129325/462161d1-854e-4199-bce3-c03ee0ab376f


src/apps/backend/modules/account/internal/account-reader.ts
```diff
@@ -39,6 +37,18 @@ export default class AccountReader {
     return account;
   }
 
+  public static async getAccountById(accountId: string): Promise<Account> {
+    const accDb = await AccountRepository.findOne({
```
codVenturer, line 41: name it accountDb

src/apps/frontend/app.component.tsx
```diff
@@ -20,15 +18,10 @@ export default function App(): React.ReactElement {
 
   return (
     <AuthProvider>
-      <BaseWebProvider>
-        <ToasterContainer
-          placement={PLACEMENT.topRight}
-          autoHideDuration={constant.TOASTER_AUTO_HIDE_DURATION}
-        />
+        <Toaster />
```
codVenturer, line 21: why this change?

balwant-jtc:
@bbbneo333 @jjalan :
I updated the drop-down icon next to the user profile to align with the theme's styling based on my recent conversation with Jai.

- Remaining AI:
1) Sidebar
2) Todo section, complete with a todo list and creation flow. 
These tasks needs to be handled separately.

jjalan:
> @bbbneo333 @jjalan : I updated the drop-down icon next to the user profile to align with the theme's styling based on my recent conversation with Jai.
> 
> * Remaining AI:
> 
> 1. Sidebar
> 2. Todo section, complete with a todo list and creation flow.
>    These tasks needs to be handled separately.

Maybe the video is not upto date, but the profile icon on navbar on the right still does not look right. It also has stripped background, plus its black. We should ensure that the name and logo look and feel matches tailwind demo - 

Screenshot 2024-02-23 at 5 00 28 PM

src/apps/frontend/components/header/dropdown-user.component.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+        <img
+          className="hidden fill-current opacity-50 sm:block"
+          src="/assets/img/icon/drop-down-arrow.svg"
+          alt="dropdown icon"
+        />
+      </Link>
+
+      {/* <!-- Dropdown Start --> */}
+      <div
```
bbbneo333, line 73: I see we have a possibility here to define a separate component for `Dropdown`. We can reuse this logic.

KrishnaVAMSI-2003:
> > @bbbneo333 @jjalan : I updated the drop-down icon next to the user profile to align with the theme's styling based on my recent conversation with Jai.
> > 
> > * Remaining AI:
> > 
> > 
> > 1. Sidebar
> > 2. Todo section, complete with a todo list and creation flow.
> >    These tasks needs to be handled separately.
> 
> Maybe the video is not upto date, but the profile icon on navbar on the right still does not look right. It also has stripped background, plus its black. We should ensure that the name and logo look and feel matches tailwind demo -
> 
> Screenshot 2024-02-23 at 5 00 28 PM

updated sir, please take a look and let me know if any changes needed

src/apps/frontend/components/header/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const { logout } = useAuthContext();
+  const { accountDetails, getAccountDetails } = useAccountContext();
+
+  const handleSidebarToggle = (state: boolean) => {
+    setIsSidebarOpen(state);
+  };
+
+  const handleSignOut = () => {
+    logout();
+    navigate('/login');
```
jjalan, line 30: Move all routes to a constant file.

src/apps/frontend/components/header/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+  };
+
+  useEffect(() => {
+    (async () => {
+      await getAccountDetails();
```
jjalan, line 35: We want to avoid component making the call and maintain state like this, have you see usage of `useAsync` where context maintains loading, loaded, error etc state for an operation? This helps to make sure data operation state is maintained in the context itself.

src/apps/frontend/components/header/user-menu-dropdown.component.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+
+type UserMenuDropdownProps = {
+  dropdownOpen: boolean;
+  dropdownRef: React.RefObject<HTMLDivElement>;
+  logout: () => void;
+  setDropdownOpen: (dropdownOpen: boolean) => void;
+};
```
jjalan, line 8: I am not sure how Tailwind admin is setup, but we want a setup like `SidebarMenuItem` where more items can be without chaging the dropdown markup or hardcoding about one particular type of dropdown menu item

jjalan (commented):
There is lot of react code which I have not looked closely, for ex- FormControl, HamburgerToggleButton, UserProfileSnippet.

Are they coming from Tailwind Admin or have we written this?

KrishnaVAMSI-2003:
> There is lot of react code which I have not looked closely, for ex- FormControl, HamburgerToggleButton, UserProfileSnippet.
> 
> Are they coming from Tailwind Admin or have we written this?

@jjalan sir, for `HamburgerToggleButton` and `UserProfileSnippet` the code is abstracted from TailAdmin only. In TailAdmin the same code written inline, but here I have created separate components. And for form-control TailAdmin there is no readily available form-control component to handle errors like in Baseweb. So created a component that accepts input component as children and displays an error(validated by Formik) line under the input.

src/apps/frontend/components/header/index.tsx
KrishnaVAMSI-2003: So sir, should I invoke the get method in the context itself. As I have seen in the `useAsync` hook, it does provides states like `isLoading`, `error` and `result` but for the method it is just providing the `asyncCallback` and we do need to invoke this method somewhere to make the api call and get required data. If we want to avoid get methods being called in a component then either we need to invoke them in either context(using useEffect) or in the `useAsync` hook. What will be the preferred?

jjalan: Its fine to invoke from component, but I want to make sure we are using the state from the context - like isloading, iserror etc instead of catch / then to render the UI diferently as the operation progress / completes

KrishnaVAMSI-2003: okay sir, I will refractor accordingly

jjalan (approved):
Please merge after addressing:

- Move api call to fetch profile from header to protected app
- Fix sidebar menu items spacing
- Sign in form should be disabled when we are in process of signing in
- Clicking on tasks shows Page not found

PR #109, opened 2024-03-05
Add Signup page to create a new account using Tailwind template

KrishnaVAMSI-2003:
## Description
This PR introduces the signup page that have been created using tailwind to create a new account.

## Database schema changes
- NA

- NA

> try logging with non existing account or creating a new account with an existing mail throws error as expected

https://github.com/jalantechnologies/boilerplate-mern/assets/74129325/b802a01d-0227-4522-a714-b1c80e4fc6f9


src/apps/frontend/contexts/auth.provider.tsx
```diff
@@ -37,6 +41,17 @@ const loginFn = async (
   return result;
 };
 
+const signupFn = async (
+  username: string,
+  password: string,
+): Promise<ApiResponse<AccessToken>> => {
+  const result = await authService.signup(username, password);
+  if (result.data) {
+    localStorage.setItem('access-token', JSON.stringify(result.data));
```
jjalan, line 50: I don't think we want to do this as part of sign up process. User should be redirected to sign in page (which you are doing) but we should not store anything in local storage.

src/apps/frontend/pages/authentication/signup/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { AsyncError } from '../../../types';
+import AuthenticationFormLayout from '../authentication-form-layout';
+import AuthenticationPageLayout from '../authentication-page-layout';
+
+import SignupForm from './signup-form';
+
+export const Signup: React.FC = () => {
+  const navigate = useNavigate();
+  const onSuccess = () => {
+    toast.success('Signup successful! Please login to continue.');
```
jjalan, line 15: Your account has been successfully created. Please login to continue.

src/apps/frontend/services/auth.service.ts
```diff
@@ -10,4 +10,12 @@ export default class AuthService extends APIService {
     username,
     password,
   });
+
+  signup = async (
+    username: string,
+    password: string,
+  ): Promise<ApiResponse<AccessToken>> => this.apiClient.post('/accounts', {
```
jjalan, line 17: I am suprised backend is sending access token on account creation. It should not.

We don't need to change backend, but we want to make sure this function should not return anything

jjalan (approved):
Please feel free to merge after addressing the comments.

Can you also attach a video of mobile experience to ensure this works well on a smaller device?

src/apps/frontend/services/auth.service.ts
KrishnaVAMSI-2003: okay sir

jjalan (commented):
We want to add re-type password functionality as well, as in here - https://demo.tailadmin.com/signup

I am not sure if our account model supports name to be set as well, if yes, we need to accept the name as well as part of sign up process

KrishnaVAMSI-2003:
> We want to add re-type password functionality as well, as in here - https://demo.tailadmin.com/signup
> 
> I am not sure if our account model supports name to be set as well, if yes, we need to accept the name as well as part of sign up process

Sir, the name field doesn't exists in the account schema. Pleae let me know if needed I will add it

jjalan:
> > I am not sure if our account model supports name to be set as well, if yes, we need to accept the name as well as part of sign up process
> 
> Sir, the name field doesn't exists in the account schema. Please let me know if needed I will add it

IMO, it would be good to add two fields - first name (required), last name (optional). In terms of layout, we can show two input box adjacent to each other instead of Full name (as per Tailwind Admin sign up page)

jjalan:
I also compared the experience on sign up page on us vs Tailwind Admin, see the difference, specifically - the background color of the page, with the form in white box, as well as, good spacing around it (where us our page has little horizontal padding around the form)

Screenshot 2024-03-05 at 7 25 08 PM

It is possible that the same problem exists with sign in page, we should compare that too and ensure its consistent.

src/apps/frontend/constants/index.ts
```diff
@@ -1,6 +1,7 @@
 const constant = {
   EMAIL_VALIDATION_ERROR: 'Please enter a valid email',
   PASSWORD_MIN_LENGTH: 8,
+  PASSWORD_MATCH_VALIDATION_ERROR: 'Passwords do not match',
```
jjalan, line 4: I understand ideally these should be provided by PM. Since there is a lack of it, I would recommend to Google - `How to write good error messages` (and or similar Google search) and see if we can self improve this.

src/apps/frontend/pages/authentication/login/login-form.tsx
```diff
@@ -20,91 +22,95 @@ const LoginForm: React.FC<LoginFormProps> = ({ onError, onSuccess }) => {
...
-      </div>
-    </div>
+        <button
+          className={`w-full cursor-pointer rounded-lg border border-primary bg-primary p-4 font-medium text-white transition hover:bg-primary/90
+            ${isLoginLoading && 'cursor-not-allowed bg-primary/70'}`
+          }
+          disabled={isLoginLoading}
+          type="submit"
+        >
+          {isLoginLoading ? 'Loading...' : 'Sign In'}
```
jjalan, line 104: Have we checked Tailwind in general? What is their treatment when for async operation?

This may give some idea - https://tailwindcss.com/docs/hover-focus-and-other-states

KrishnaVAMSI-2003:
Password reset flow will be implemented in another PR

src/apps/backend/modules/account/internal/store/account-db.ts
```diff
@@ -13,6 +15,13 @@ export const AccountDbSchema: Schema = new Schema<AccountDB>(
       type: Boolean,
       required: true,
     },
+    firstName: {
+      type: String,
+      required: true,
+    },
+    lastName: {
+      type: String,
```
jjalan, line 23: After I gave you the feedback on keeping it optional, I thought a bit more about it. I think we want to make last name as required as well.

src/apps/frontend/components/header/user-profile-snippet.component.tsx
```diff
@@ -22,7 +22,10 @@ const UserProfileSnippet: React.FC<DropdownUserProps> = ({
   const trigger = useRef<HTMLAnchorElement>(null);
   const dropdown = useRef<HTMLDivElement>(null);
 
-  const username = account?.name || account?.username || 'Guest';
+  let username = account?.username || 'Guest';
+  if (account?.firstName) {
+    username = `${account?.firstName} ${account?.lastName}`;
+  }
```
jjalan, line 28: I don't understand why account would be nil here. I would expect the user profile snippet to show only if there is an account.

Moving forward, with this change, I expect first and last name to be available so there is not a whole lot of need to do if else here.

I would also add `displayName` in `Account` object on client side which does this concatenation to better encapsulate the logic

src/apps/frontend/pages/authentication/login/login-form-checkbox.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+
+const LoginFormCheckbox: React.FC = () => (
+    <div className="relative pt-0.5">
```
jjalan, line 4: Can you check if this is 2 spaces instead of 4?

src/apps/backend/modules/account/internal/store/account-db.ts
KrishnaVAMSI-2003: its minimum length will be 1 and error message will be `Last name should not be empty` is it fine sir?

src/apps/frontend/constants/index.ts
```diff
@@ -1,7 +1,9 @@
 const constant = {
   EMAIL_VALIDATION_ERROR: 'Please enter a valid email',
-  FIRST_NAME_MIN_LENGTH: 2,
-  FIRST_NAME_VALIDATION_ERROR: 'Please enter at least 2 character long first name',
+  FIRST_NAME_MIN_LENGTH: 1,
+  FIRST_NAME_VALIDATION_ERROR: 'First name should not be empty',
```
jjalan, line 4: Please specify our first name

src/apps/frontend/constants/index.ts
```diff
@@ -1,7 +1,9 @@
...
+  LAST_NAME_MIN_LENGTH: 1,
+  LAST_NAME_VALIDATION_ERROR: 'Last name should not be empty',
```
jjalan, line 6: Please specify your last name

src/apps/frontend/types/account.ts
```diff
@@ -2,12 +2,14 @@ import { JsonObject } from './common-types';
 
 export class Account {
   id: string;
+  displayName: string;
```
jjalan, line 5: Instead of storing in property and computing in constructor, consider exposing `displayName = () => {}` method on the class.

jjalan:
In the video, the header of the form Sign In - font size seems smaller than Tailwind Admin. Similarly I don't think Tailwind Admin has `hr`. Can you please make sure that form experience aligns with Tailwind?

jjalan (approved):

PR #116, opened 2024-03-07
[WIP] Add reset password flow in the app using tailwind template

KrishnaVAMSI-2003:
## Description
This PR introduces the password reset feature, clicking the forgot password text in the login page will redirects to the forgot password page in which user will enter their email address and can request for reset link through email.

## Database schema changes
- added a new password reset token schema

```
{
  account: {
    ref: 'Account',
    required: true,
    type: Schema.Types.ObjectId,
  },
  expiresAt: { type: Date, required: true },
  token: { type: String, required: true },
  isUsed: {
    type: Boolean,
    default: false,
  },
},
{
  timestamps: {
    createdAt: 'createdAt',
    updatedAt: 'updatedAt',
  },
}
```

- POST /password-reset-tokens
  - should create a new password reset token and send an email with password reset link, if enabled for the env
  - should throw an error if account with username does not exists
- PATCH /account/:accountId/reset-password
  - should reset the account password to the new password
  - should throw 404 if account does not exist
  - should throw 404 if the account's password reset token is not found
  - should throw error if the password reset token is already used
  - should throw error if the password reset token does not match with the token passed in the payload
  - should throw error if the password reset token does not match with the token passed in the payload

https://github.com/jalantechnologies/boilerplate-mern/assets/74129325/ec040203-072d-4a3c-a006-ccffae62ff54


src/apps/backend/modules/account/account-service.ts
```diff
@@ -22,4 +34,81 @@ export default class AccountService {
   ): Promise<Account> {
     return AccountReader.getAccountById(params.accountId);
   }
+
+  public static async createPasswordResetToken(
```
jjalan, line 38: Just like access token, should password reset token by separate module, ex: `password-reset-token`

src/apps/backend/modules/account/account-service.ts
```diff
@@ -22,4 +34,81 @@ export default class AccountService {
...
+    await this.sendPasswordResetEmail(
+      account.id,
+      account.firstName,
+      account.username,
+      passwordResetToken,
+    );
+    return passwordResetTokenDbData;
+  }
+
+  public static async resetPassword(
```
jjalan, line 59: There should be an existing PATCH call on account, right? If not we should have that i.e we should continue to have a single path to update properties on account, include password.

KrishnaVAMSI-2003: But sir, for resetting password first we need to verify the token that have been sent to the email and the params will be different. Whereas, for other account updates like firstName/lastName we will verify the JWT token in the header using auth middleware.
The reset password route needs to be a public and different PATCH call, as we don't have any auth header because we are resetting the password without login. And I believe 
> we should continue to have a single path to update properties on account, include password.

this is possible when the user is logged in.

src/apps/frontend/pages/authentication/forgot-password/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { AsyncError } from '../../../types';
+import useTimer from '../../../utils/use-timer';
+import AuthenticationFormLayout from '../authentication-form-layout';
+import AuthenticationPageLayout from '../authentication-page-layout';
+
+import ForgotPasswordForm from './forgot-password-form';
+import ForgotPasswordResendEmail from './forgot-password-resend-email';
+
+export const ForgotPassword: React.FC = () => {
+  const [isResendEmailPage, setIsResendEmailPage] = useState<boolean>(false);
```
tara122333, line 14: No need to declare state type if we initialize the state value
`const [isResendEmailPage, setIsResendEmailPage] = useState(false);` --> now state type is boolean
`const [username, setUsername] = useState('');` --> now state type is string

src/apps/frontend/pages/authentication/reset-password/reset-password-form.hook.ts
```diff
@@ -0,0 +1,59 @@
...
+  const token = new URLSearchParams(search).get('token');
+
+  const {
+    isResetPasswordLoading, resetPasswordError, resetPassword,
+  } = useResetPasswordContext();
+
+  const formik = useFormik({
+    initialValues: {
+      password: '',
+      retypePassword: '',
```
tara122333, line 27: Can we use `confirmPassword` instead of `retypePassword` ?

src/apps/frontend/pages/authentication/reset-password/reset-password-form.hook.ts
```diff
@@ -0,0 +1,59 @@
...
+        .required(constant.PASSWORD_VALIDATION_ERROR),
+      retypePassword: Yup.string()
+        .oneOf([Yup.ref('password')], constant.PASSWORD_MATCH_VALIDATION_ERROR)
+        .required(constant.PASSWORD_MATCH_VALIDATION_ERROR),
+    }),
+    onSubmit: (values) => {
+      resetPassword(
+        accountId,
+        values.password,
+        token,
```
tara122333, line 41: We should define a type so we don't have to keep passing accountId, password and token (as string) separately.

jjalan:
For a UI that has async operation, we should ensure we handle loading state i.e. when operation is in progress. Form, button should be disabled and user should see some sort of loading icon

src/apps/backend/modules/password-reset-token/password-reset-token-service.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  private static async sendPasswordResetEmail(
+    accountId: string,
+    firstName: string,
+    username: string,
+    passwordResetToken: string,
+  ): Promise<void> {
+    const passwordResetEmailEnabled = JSON.parse(ConfigService.getValue<string>(
+      'accounts.passwordResetEmailEnabled',
+    )) as boolean;
```
jjalan, line 88: This is going to try to typecast string `PASSWORD_RESET_EMAIL_ENABLED` to boolean which seems wrong also. Don't you have `ConfigService.getValue<boolean>`?

src/apps/frontend/pages/authentication/forgot-password/back-button.tsx
```diff
@@ -0,0 +1,18 @@
...
+interface BackButtonProps {
+  onClick: () => void;
+}
+
+const BackButton: React.FC<BackButtonProps> = ({
+  onClick,
+}) => (
+    <button
+      onClick={onClick}
+      className="mb-5 cursor-pointer text-lg transition active:text-primary/80"
```
jjalan, line 12: `mb-5`: Its not the job of the component to add margins around it, but rather the container to layout the components and put spacing between them.

src/apps/backend/modules/account/rest-api/account-router.ts
```diff
@@ -10,6 +10,8 @@ export default class AccountRouter extends ApplicationRouter {
 
     router.post('/', ctrl.createAccount);
 
+    router.patch('/:accountId/reset-password', ctrl.resetPassword);
```
jjalan, line 13: This is not a CRUD.

Can we just keep it PATCH? We allow updating password when user is not authenticated

src/apps/frontend/pages/authentication/reset-password/reset-password-form.tsx
```diff
@@ -0,0 +1,130 @@
...
+              disabled={isResetPasswordLoading}
+              error={formik.touched.password && formik.errors.password}
+              name="password"
+              onBlur={formik.handleBlur}
+              onChange={formik.handleChange}
+              type={isPasswordVisible ? 'text' : 'password'}
+              value={formik.values.password}
+              placeholder="Enter your new password"
+            />
+            <button
```
jjalan, line 57: use Button component

src/apps/frontend/pages/authentication/reset-password/reset-password-form.tsx
```diff
@@ -0,0 +1,130 @@
...
+const ResetPasswordForm: React.FC<ResetPasswordFormProps> = ({
+  onSuccess,
+  onError,
+}) => {
+  const { formik, isResetPasswordLoading } = useResetPasswordForm({ onSuccess, onError });
+
+  const [isPasswordVisible, setIsPasswordVisible] = useState(false);
+  const [isRetypePasswordVisible, setIsRetypePasswordVisible] = useState(false);
+
+  function togglePasswordVisibility() {
```
jjalan, line 23: why ES5 syntax?

src/apps/frontend/components/button/button-primary.tsx
jjalan, line 1: rename file to button

src/apps/frontend/pages/authentication/reset-password/reset-password-form.tsx
```diff
@@ -0,0 +1,130 @@
...
+                  alt="hide password icon"
+                />
+              ) : (
+                <img
+                  className="size-6.5 opacity-65"
+                  src="/assets/img/icon/eye-open.svg"
+                  alt="show password icon"
+                />
+              )}
+            </button>
```
jjalan, line 116: sounds like we need to have password input that toggle password visibility. I see this being repeated

src/apps/frontend/pages/authentication/forgot-password/forgot-password-form.tsx
```diff
@@ -0,0 +1,73 @@
...
+  onError,
+  onSuccess,
+}) => {
+  const navigate = useNavigate();
+  const isLoginLoading = false;
+  const { formik } = useForgotPasswordForm({ onError, onSuccess });
+  return (
+    <>
+      <BackButton onClick={() => navigate(constants.LOGIN)} />
+      <h2 className="mb-5 text-2xl font-bold text-black dark:text-white sm:text-title-xl2">
```
jjalan, line 27: Can we create `H2` component? I see similar css applied on H2 multiple places. The component should not have `mb5` as the layout should control spacing around

src/apps/frontend/pages/authentication/forgot-password/forgot-password-form.tsx
```diff
@@ -0,0 +1,73 @@
...
+        Forgot Password?
+      </h2>
+      <p className="mb-4 text-xl font-medium">
```
jjalan, line 30: Maybe `ParagraphMedium` (without `mb-4`)

src/apps/frontend/components/button/button-primary.tsx
```diff
@@ -0,0 +1,37 @@
...
+  type,
+}) => (
+  <button
+    className={`w-full cursor-pointer rounded-lg border border-primary bg-primary p-4 font-medium text-white transition active:bg-primary/80
+      ${(disabled || isLoading) ? 'cursor-not-allowed bg-primary/80' : 'hover:bg-primary/90'}`}
+    disabled={disabled || isLoading}
+    type={type}
+    onClick={onClick}
+  >
+  {isLoading ? <Spinner /> : label}
```
tara122333, line 33: create the component with `PropsWithChildren` and use `children` instead of `label`

src/apps/frontend/pages/authentication/forgot-password/back-button.tsx
tara122333, line 1: Create this file in the `component` folder. that we can use in other pages/components.

jjalan: In an ideal, this should not be even a component and we should just inline in the page

PR #118, opened 2024-03-10
Implement Task List Component 

aatiqJalanSWE:
## Description
The pull request introduces the Task List component, enhancing user experience and navigation within the application so that users can create, update, and delete tasks according to their needs.

Also added title and  description entity in TaskDbSchema.

The main features include:

### TaskHeader Enhancement:

```
Integrated the TaskHeader component for managing tasks, providing a cleaner interface.

Improved styling and organization of the task-related content.
```

### Task Display:

```
Designed a task display section with a clear structure.

```

ScreenRecording:

https://github.com/jalantechnologies/boilerplate-mern/assets/126680323/c78d5955-2d3a-4623-9346-a30e96e10799

Recording of login flow to task list UI

https://github.com/jalantechnologies/boilerplate-mern/assets/126680323/9b88ed8b-2db1-4a35-be9f-b0c6f3fc9aa1

## Database schema changes
_Please document any change in database schema made part of this PR._

TaskDbSchema

```
description: {
      type: String,
      index: true,
      required: true,
    },
    title: {
      type: String,
      index: true,
      required: true,
    },
```

PATCH /tasks/:id/updateTask

should throw 404 if taskId does not exist
should throw 404 if account does not exist

https://github.com/jalantechnologies/boilerplate-mern/assets/126680323/94716d9e-25ba-4596-848f-7135beae534a


src/apps/backend/modules/task/internal/task-reader.ts
```diff
@@ -25,14 +25,14 @@ export default class TaskReader {
     return TaskUtil.convertTaskDBToTask(taskDb);
   }
 
-  public static async getTaskByNameForAccount(params: GetTaskByNameParams): Promise<Task> {
+  public static async getTaskByTitleForAccount(params: GetTaskByTitleParams): Promise<Task> {
```
jjalan, line 28: This does not sound right, task title cannot guaranteed be unique. For example, I can add a daily task to work out.

src/apps/backend/modules/task/internal/task-writer.ts
```diff
@@ -13,16 +13,17 @@ export default class TaskWriter {
   public static async createTask(params: CreateTaskParams): Promise<Task> {
     const existingTask = await TaskRepository.findOne({
```
jjalan, line 14: Remove this, and error below

src/apps/frontend/components/breadcrumb/breadcrumb.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import { Link } from 'react-router-dom';
+
+interface BreadcrumbProps {
+  pageName: string;
+}
+const Breadcrumb: React.FC<BreadcrumbProps> = ({ pageName }) => (
```
jjalan, line 7: add new line above.

jjalan: Is this implementation coming from Tailwind admin? If yes, then it may be ok, but as I see it has few limitation:

1. Breadcrumb can have multiple levels, this component only supports 1
2. Ideally there should be breadcrumb and breadcrumb items
3. `pageName` seems like a concept related to page, this should be more likely breadcrumbItem or something like that
4. Is it `bread-crumb` or `breadcrumb` i.e. are those 2 words or one word?

src/apps/frontend/components/dropdowns/dropdown-default.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const [dropdownOpen, setDropdownOpen] = useState(false);
+  const [popupOpen, setPopupOpen] = useState(false);
+
+  return (
+    <div className="relative flex">
+      <button
+        className="text-[#98A6AD] hover:text-body"
+        onClick={() => setDropdownOpen(!dropdownOpen)}
+      >
+        <svg
```
jjalan, line 15: Instead of inlining, move it icons folder.

I also think there is already a dropdown component created in the app and `dropdown-default` is not a good component name, not sure why we have two anyways

src/apps/frontend/components/task/task-popup.tsx
```diff
@@ -0,0 +1,96 @@
+import React, { useState } from 'react';
+
+interface TaskPopupProps {
+  popupOpen: boolean;
+  setPopupOpen: (open: boolean) => void;
+}
+
+const TaskPopup: React.FC<TaskPopupProps> = ({ popupOpen, setPopupOpen }) => {
```
jjalan, line 8: I think this is TaskModal, please see difference between modal and popup.

Also, is this copied from TailwinAdmin, there is things that should be improved in the React code

jjalan (commented):
Please verify that the experience matches that of Tailwind Admin template

src/apps/frontend/components/task/task-popup.tsx
aatiqJalanSWE: I will correct this to TaskModal.

src/apps/frontend/components/dropdowns/dropdown-default.tsx
```diff
@@ -0,0 +1,104 @@
...
+            d="M15.75 11.25C16.9926 11.25 18 10.2426 18 9C18 7.75736 16.9926 6.75 15.75 6.75C14.5074 6.75 13.5 7.75736 13.5 9C13.5 10.2426 14.5074 11.25 15.75 11.25Z"
+            fill=""
+          />
+        </svg>
+      </button>
+      <div
+        onFocus={() => setDropdownOpen(true)}
+        onBlur={() => setDropdownOpen(false)}
+        className={`absolute right-0 top-full z-40 w-40 space-y-1 rounded-sm border border-stroke bg-white p-1.5 shadow-default dark:border-strokedark dark:bg-boxdark ${
+          dropdownOpen === true ? 'block' : 'hidden'
```
tara122333, line 41: can we use `dropdownOpen ? 'block' : 'hidden'` instead of `dropdownOpen === true ? 'block' : 'hidden'` ?

src/apps/frontend/components/task/task-popup.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
```
tara122333, line 10: Can we use formik for handling form input values?

src/apps/backend/modules/task/internal/task-writer.ts
```diff
@@ -11,23 +11,38 @@ import TaskUtil from './task-util';
...
+        $set: {
+          description: params.description,
+          title: params.title,
+        },
+      },
+      { new: true },
+    );
+
+    if (!task) {
+      throw new Error('Task not found');
```
jjalan, line 40: Please improve copyright message. There is lot of literature online about, and if you Google `how to write good error messages`, you can find some reference.

Also, I think we need to throw with appropriate HTTP status code. For example, this should probably throw 404 in case there is no task found. You may need to create a custom error message from here instead of Error

src/apps/backend/modules/task/types.ts
```diff
@@ -18,14 +19,22 @@ export type GetTaskParams = {
   taskId: string;
 };
 
-export type GetTaskByNameParams = {
-  accountId: string,
-  name: string;
+export type GetTaskByTitleParams = {
```
jjalan, line 22: This is not being used, please remove

src/apps/backend/modules/task/types.ts
```diff
@@ -43,11 +52,11 @@ export enum TaskErrorCode {
   TASK_ALREADY_EXISTS = 'TASK_ERR_02',
 }
 
-export class TaskWithNameExistsError extends ApplicationError {
+export class TaskWithTitleExistsError extends ApplicationError {
```
jjalan, line 55: This should not be used, as title is not unique

src/apps/backend/modules/task/types.ts
```diff
@@ -63,11 +72,11 @@ export class TaskNotFoundError extends ApplicationError {
   }
 }
 
-export class TaskWithNameNotFoundError extends ApplicationError {
+export class TaskWithTitleNotFoundError extends ApplicationError {
```
jjalan, line 75: This should not be used, as title is not unique

src/apps/frontend/app.component.tsx
```diff
@@ -19,10 +19,12 @@ export default function App(): React.ReactElement {
   return (
     <AuthProvider>
       <AccountProvider>
-        <Toaster />
-        <Router>
-          <AppRoutes />
-        </Router>
+        <TaskProvider>
```
jjalan, line 22: Why does all the routes needs to be within TaskProvider, only Task specific routes should have this context.

src/apps/frontend/components/task/task-list.tsx
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+
+import BreadcrumbItem from '../breadcrumb/breadcrumb-item';
+import EllipsisMenu from '../ellipsis/ellipsis-menu';
+
+import TaskHeader from './task-header';
+
+const TaskList: React.FC = () => (
+    <div className="p-5 sm:p-10">
```
jjalan, line 9: Use 2 spaces instead of 4 spaces

src/apps/frontend/components/task/task-modal.tsx
```diff
@@ -0,0 +1,68 @@
...
+          className="absolute right-1 top-1 sm:right-5 sm:top-5"
+        >
+          <img src="assets/svg/close-icon.svg" className="fill-current" />
+        </button>
+
+        <div className="my-5">
+          <div className="mb-2.5 block font-medium text-black dark:text-white">
+            {title}
+          </div>
+          <input
```
jjalan, line 35: We probably want to use formik

src/apps/frontend/components/ellipsis/ellipsis-menu.tsx
```diff
@@ -0,0 +1,45 @@
+import React, { useState } from 'react';
+
+import TaskModal from '../task/task-modal';
+
+const EllipsisMenu: React.FC = () => {
```
jjalan, line 5: This is VERY poor abstraction.

You have a generic component inside `components` call ellipsis menu which refers task.

The component name is also not clear

src/apps/frontend/components/breadcrumb/breadcrumb-item.tsx
```diff
@@ -0,0 +1,26 @@
...
+const BreadcrumbItem: React.FC<BreadcrumbItemProps> = ({ pageName }) => (
+  <div className="mb-6 flex flex-col gap-3 sm:flex-row sm:items-center sm:justify-between">
+    <div className="text-title-md2 font-semibold text-black dark:text-white">
+      {pageName}
+    </div>
+    <div>
+      <div className="flex items-center gap-2">
+        <div>
+          <Link className="font-medium" to="/">
+            Dashboard /
```
jjalan, line 17: Why would this belong to breadcrumb item?

I would think that that Breadcrumb takes a list of breadcrumb item (ReactNode type) and assemble them dynamically. The breadcrumb component is responsible to use `Dashboard` at the top breadcrumb item

src/apps/frontend/components/index.ts
```diff
@@ -1,5 +1,10 @@
+import BreadcrumbItem from './breadcrumb/breadcrumb-item';
+import EllipsisMenu from './ellipsis/ellipsis-menu';
 import FormControl from './form-control';
 import Header from './header';
 import Input from './input';
+import TaskList from './task/task-list';
```
jjalan, line 6: Task list is not generic component, should not be here

src/apps/frontend/pages/tasks/index.tsx
```diff
@@ -1,5 +1,11 @@
 import * as React from 'react';
 
-const Tasks: React.FC = () => <div></div>;
+import { TaskList } from '../../components';
+
+const Tasks: React.FC = () => (
+  <div>
```
jjalan, line 6: Why do we need `div`?

src/apps/frontend/services/task.service.ts
```diff
@@ -0,0 +1,42 @@
...
+  };
+
+  updateTask = async (
+    taskId: string,
+    taskData: Partial<Task>,
+  ): Promise<ApiResponse<Task>> => {
+    const userAccessToken = JSON.parse(
+      localStorage.getItem('access-token'),
+    ) as AccessToken;
+    return this.apiClient.put(`/tasks/${taskId}`, taskData, {
```
jjalan, line 25: I think the concept of JSON object ans JS object is still not clear.

Server returns a JSON , say `{date: '2020-12-20'}`. The job of the service is to convert that into Javascript object `new Task(json)`. This is true in `getTasks`

src/apps/backend/modules/task/internal/store/task-db.ts
```diff
@@ -20,7 +21,12 @@ export const TaskDbSchema: Schema = new Schema<TaskDB>(
       index: true,
       required: true,
     },
-    name: {
+    description: {
+      type: String,
+      index: true,
```
dev-rathore, line 26: `index: true` is not required if we do not have any search by description reader method.

src/apps/backend/modules/task/rest-api/task-router.ts
```diff
@@ -13,6 +13,7 @@ export default class TaskRouter extends ApplicationRouter {
     router.post('/', ctrl.createTask);
     router.get('/', ctrl.getTasks);
     router.get('/:id', ctrl.getTask);
+    router.put('/:id', ctrl.updateTask);
```
dev-rathore, line 16: Replace `put` with `patch`

src/apps/frontend/types/task.ts
```diff
@@ -0,0 +1,24 @@
+import { JsonObject } from './common-types';
+
+export interface UpdateTaskParams {
+  accountId: string;
+  description: string;
+  taskId: string;
+  title: string;
+}
+export interface DeleteTaskParams {
```
dev-rathore, line 9: Are we using this `DeleteTaskParams` interface somewhere, if not then this can be removed.

src/apps/backend/modules/task/rest-api/task-router.ts
aatiqJalanSWE: sure

src/apps/backend/modules/task/internal/store/task-db.ts
```diff
@@ -20,7 +21,12 @@ export const TaskDbSchema: Schema = new Schema<TaskDB>(
...
+      required: true,
+    },
+    title: {
       type: String,
       index: true,
```
KrishnaVAMSI-2003, line 31: can we remove index here too? cause no reader method is performing search by title

src/apps/frontend/components/breadcrumb/breadcrumb-item.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+const BreadcrumbItem: React.FC<BreadcrumbItemProps> = ({ pageName }) => (
+  <div className="mb-6 flex flex-col gap-3 sm:flex-row sm:items-center sm:justify-between">
+    <div className="text-title-md2 font-semibold text-black dark:text-white">
+      {pageName}
+    </div>
+    <div>
+      <div className="flex items-center gap-2">
+        <div>
+          <Link className="font-medium" to="/">
```
KrishnaVAMSI-2003, line 16: use route `DASHBOARD` of constants here instead of direct path `/`

src/apps/frontend/components/task/task-modal.tsx
```diff
@@ -0,0 +1,68 @@
...
+const TaskModal: React.FC<TaskModalProps> = ({
+  modalOpen,
+  setModalOpen,
+  btnText,
+}) => {
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+
+  return (
+    <div
```
KrishnaVAMSI-2003, line 18: can we create a global modal component and use it here, to make it reusable

src/apps/backend/modules/task/internal/store/task-db.ts
aatiqJalanSWE: done with this in upcoming PR

src/apps/frontend/components/ellipsis/ellipsis-menu.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const [dropdownOpen, setDropdownOpen] = useState(false);
+  const [modalOpen, setModalOpen] = useState(false);
+
+  return (
+    <div className="relative flex">
+      <button
+        className="text-[#98A6AD] hover:text-body"
```
tara122333, line 12: `text-[#98A6AD]` is not correct. Please declare the Color code in `tailwind.config.cjs`

src/apps/frontend/components/task/task-header.tsx
```diff
@@ -0,0 +1,31 @@
+import React, { useState } from 'react';
+
+import TaskModal from './task-modal';
+
+const TaskHeader: React.FC = () => {
+  const [modalOpen, setModalOpen] = useState(false);
```
tara122333, line 6: Can we use `isModalOpen` instead of `modalOpen`?

src/apps/frontend/components/task/task-list.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+              <div className="flex flex-col gap-2">
+                <label htmlFor="taskCheckbox1" className="cursor-pointer">
+                  <div className="relative flex items-center pt-0.5">
+                    <input
+                      type="checkbox"
+                      id="taskCheckbox1"
+                      className="taskCheckbox sr-only"
+                    />
+                    <div className="box mr-3 flex size-5 items-center justify-center rounded border border-stroke dark:border-strokedark dark:bg-boxdark-2">
```
tara122333, line 32: Why are you using custom className `box`?

src/apps/frontend/components/task/task-modal.tsx
```diff
@@ -0,0 +1,68 @@
...
+          </div>
+          <textarea
+            cols={30}
+            rows={7}
+            placeholder="Enter task description"
+            className="w-full rounded-sm border border-stroke bg-white px-4.5 py-3 text-black focus:border-primary focus-visible:outline-none dark:border-strokedark dark:bg-boxdark dark:text-white dark:focus:border-primary"
+            onChange={(e: React.ChangeEvent<HTMLTextAreaElement>) => {
+              setDescription(e.target.value);
+            }}
+          ></textarea>
```
tara122333, line 57: Create a separate component

src/apps/backend/modules/task/task-service.ts
```diff
@@ -18,12 +18,12 @@ export default class TaskService {
...
 
-  public static async getTaskForAccount(params: GetTaskParams): Promise<Task> {
-    return TaskReader.getTaskForAccount(params);
+  public static async updateTask(params: UpdateTaskParams): Promise<Task> {
+    return TaskWriter.updateTask(params);
   }
 
-  public static async getTaskByNameForAccount(params: GetTaskByNameParams): Promise<Task> {
-    return TaskReader.getTaskByNameForAccount(params);
+  public static async getTaskForAccount(params: GetTaskParams): Promise<Task> {
```
jjalan, line 25: There is potentially wrong with the function name.

An account can have multiple task, right?

src/apps/frontend/components/layouts/task-item.layout.tsx
```diff
@@ -0,0 +1,9 @@
+import React, { PropsWithChildren } from 'react';
+
+const TaskItemLayout: React.FC<PropsWithChildren> = ({ children }) => (
```
jjalan, line 3: Why is TaskItem a common component i.e. why is it under `components/layouts` folder?

src/apps/frontend/components/task/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+import HeadingLarge from '../typography/heading-large';
+
+import TaskHeader from './task-header';
+import TaskSection from './task-section';
+
+interface TaskContainerProps {
+  onError: (error: AsyncError) => void;
+}
+
+const TaskContainer: React.FC<TaskContainerProps> = ({ onError }) => {
```
jjalan, line 15: Why is TaskContainer a common component i.e. why is it under components folder?

src/apps/frontend/components/task/task-header.tsx
```diff
@@ -0,0 +1,46 @@
...
+import toast from 'react-hot-toast';
+
+import useTaskForm from '../../pages/tasks/tasks-form.hook';
+import { TaskOperationType } from '../../types/task';
+import Button from '../button';
+import TaskItemLayout from '../layouts/task-item.layout';
+
+import TaskModal from './task-modal';
+
+const TaskHeader: React.FC = () => {
```
jjalan, line 11: Why is TaskHeader a common component i.e. why is it under components folder?

src/apps/frontend/components/task/task-modal.tsx
```diff
@@ -0,0 +1,107 @@
...
+  task?: Task;
+  formik: FormikProps<Task>;
+  setFieldValue: (
+    formik: FormikProps<Task>,
+    field: string,
+    value: string,
+  ) => void;
+}
+
+const TaskModal: React.FC<TaskModalProps> = ({
```
jjalan, line 24: Why is TaskModal a common component i.e. why is it under components folder?

src/apps/frontend/components/task/task-section.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+import TaskModal from './task-modal';
+
+interface TaskBlockSectionProps {
+  tasks: Task[];
+  handleDeleteTask: (taskId: string) => void;
+  isGetTasksLoading: boolean;
+}
+
+const TaskSection: React.FC<TaskBlockSectionProps> = ({
```
jjalan, line 22: Why is TaskSection a common component i.e. why is it under components folder?

src/apps/frontend/components/typography/heading-large.tsx
```diff
@@ -0,0 +1,11 @@
+import React, { PropsWithChildren } from 'react';
+
+const HeadingLarge: React.FC<PropsWithChildren> = ({
+  children,
+}) => (
+  <p className="text-3xl font-semibold text-black">
```
jjalan, line 6: Why are using `p` for heading and not `h` tags? This is semantically invalid HTML

src/apps/frontend/components/typography/heading-small.tsx
```diff
@@ -0,0 +1,11 @@
+import React, { PropsWithChildren } from 'react';
+
+const HeadingSmall: React.FC<PropsWithChildren> = ({
+  children,
+}) => (
+  <p className="text-2xl font-semibold text-black">
```
jjalan, line 6: Why are using p for heading and not h tags? This is semantically invalid HTML

src/apps/frontend/components/typography/label-large.tsx
```diff
@@ -0,0 +1,11 @@
+import React, { PropsWithChildren } from 'react';
+
+const LabelLarge: React.FC<PropsWithChildren> = ({
+  children,
+}) => (
+  <p className="text-lg font-medium text-black">
```
jjalan, line 6: Why are using p for heading and not h tags? This is semantically invalid HTML

src/apps/frontend/components/typography/label-medium.tsx
```diff
@@ -0,0 +1,11 @@
+import React, { PropsWithChildren } from 'react';
+
+const LabelMedium: React.FC<PropsWithChildren> = ({
+  children,
+}) => (
+  <p className="text-base font-medium text-black">
```
jjalan, line 6: Why are using p for heading and not label tags? This is semantically invalid HTML

src/apps/backend/modules/task/task-service.ts
aatiqJalanSWE: In the case of a single account, we are able to add more than one task as the title and description.

src/apps/frontend/components/layouts/task-item.layout.tsx
aatiqJalanSWE: Basically, TaskItem Layouts are used for adding multiple tasks which contain a title and description, so I made a common layout to add the tasks that fall under TaskItem Layouts.

src/apps/backend/modules/task/types.ts
jjalan: @aatiqJalanSWE - Why was this resolved?

I don't think `TaskWithTitleNotFoundError` should be valid any more as title is no longer unique and we should not be searching via title

src/apps/frontend/app.component.tsx
```diff
@@ -19,10 +19,10 @@ export default function App(): React.ReactElement {
   return (
     <AuthProvider>
       <AccountProvider>
-        <Toaster />
-        <Router>
-          <AppRoutes />
-        </Router>
+          <Toaster />
```
jjalan, line 22: Kindly undo unrelated formatting changes here and other places

src/apps/frontend/components/layouts/task-item.layout.tsx
jjalan: Why is it under `common` layout?

The way to design components are:

1. If its generic layout, then it should be named as such VeritcalStack etc.
2. If its not generic layout, it should not be in components folder

src/apps/frontend/app.component.tsx
aatiqJalanSWE: sure

src/apps/backend/modules/task/types.ts
aatiqJalanSWE: yes, fix this

src/apps/frontend/pages/tasks/task-container.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { AsyncError } from '../../types';
+
+import TaskHeader from './task-header';
+import TaskSection from './task-section';
+
+interface TaskContainerProps {
+  onError: (error: AsyncError) => void;
+}
+
+const TaskContainer: React.FC<TaskContainerProps> = ({ onError }) => {
```
jjalan, line 14: No need for this, as this is not being reused.

Inline this directly in `const Tasks: React.FC = () => {`

src/apps/frontend/pages/tasks/task-header.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  const { addTaskFormik, setFormikFieldValue } = useTaskForm({
+    onError,
+    onSuccess,
+  });
+
+  return (
+    <TaskItemCard>
+      <Flex alignItems='center' justifyContent='between'>
+        <FlexItem >
```
jjalan, line 34: extra space between `m` and `>`, please remove it

src/apps/frontend/pages/tasks/task-header.tsx
```diff
@@ -0,0 +1,56 @@
...
+          <HeadingLarge>Tasks</HeadingLarge>
+        </FlexItem>
+        <FlexItem >
```
jjalan, line 37: extra space between m and >, please remove it

src/apps/frontend/pages/tasks/task-header.tsx
```diff
@@ -0,0 +1,56 @@
...
+    setIsModalOpen(false);
+  };
+
+  const { addTaskFormik, setFormikFieldValue } = useTaskForm({
+    onError,
+    onSuccess,
+  });
+
+  return (
+    <TaskItemCard>
```
jjalan, line 32: It is not clear why `TaskHeader` which has no task item uses `TaskItemCard`. This is odd

src/apps/frontend/pages/tasks/task-modal.tsx
```diff
@@ -0,0 +1,109 @@
...
+const TaskModal: React.FC<TaskModalProps> = ({
+  btnText,
+  formik,
+  isModalOpen,
+  setFieldValue,
+  setIsModalOpen,
+  task,
+  taskOperationType,
+}) => {
+  useEffect(() => {
```
jjalan, line 38: I think I mentioned this, avoid using `useEffect`

src/apps/frontend/pages/tasks/tasks-form.hook.ts
```diff
@@ -0,0 +1,116 @@
...
+
+interface TaskFormProps {
+  onError?: (error: AsyncError) => void;
+  onSuccess?: () => void;
+}
+const useTaskForm = ({
+  onError,
+  onSuccess,
+}:
+TaskFormProps) => {
```
jjalan, line 17: Syntax or its formatting looks odd.

src/apps/frontend/routes/protected.tsx
```diff
@@ -15,7 +15,9 @@ const App = () => {
 
   return (
     <AppLayout>
-      <Outlet />
+      <TaskProvider>
```
jjalan, line 18: We should ensure that TaskProvider is only around task route(s)

jjalan (changes requested):
The checkbox in the task is not right. It was right design paradigm when there are multiple items associated with a task. The checkbox interface allows user to check as subtasks are done.

In our case, we do not have subtasks, so checkbox does not make sense.

After addressing comments, please make sure that all videos are updated in the PR description

src/apps/frontend/pages/tasks/task-container.tsx
aatiqJalanSWE: Sure

src/apps/frontend/pages/tasks/task-header.tsx
aatiqJalanSWE: TaskItemcard is essentially a background card for both the header and the task section.

src/apps/frontend/components/button/index.tsx
```diff
@@ -1,26 +1,28 @@
...
-  kind?: ButtonKind;
 }
 
 const Button: React.FC<PropsWithChildren<ButtonProps>> = ({
   children,
   disabled,
   isLoading,
+  kind = ButtonKind.PRIMARY,
   onClick,
+  size,
```
abhishek310001, line 24: Can you assign a default value to `size` just like kind and type?

aatiqJalanSWE: According to our team, the button component they created the secondary button without padding, if I apply the default size to the button, then it also reflects on the login page which is not properly align  as shown in the below image.

image

src/apps/frontend/pages/tasks/task-modal.tsx
aatiqJalanSWE: We are using the same modal for add and edit so that we can achieve reusability. useEffect will help us to check whether to open add or edit modal.
The alternative way could be different modals for add and edit if we want to avoid useEffect.

src/apps/frontend/pages/tasks/task-header.tsx
aatiqJalanSWE: Fixed by renaming the file to TaskCard.

src/apps/frontend/pages/tasks/task-modal.tsx
jjalan: That is not good approach either.

Are there other solutions?

aatiqJalanSWE: On the Modal mounting, we would like to know what type of modal it should be, i.e., "add" or "edit". However, when I tried other implementations, it was breaking our code, as it was working properly when using useEffect in the mounting of modals.

src/apps/frontend/pages/tasks/task-header.tsx
```diff
@@ -0,0 +1,56 @@
...
+    setIsModalOpen(false);
+  };
+
+  const { addTaskFormik, setFormikFieldValue } = useTaskForm({
+    onError,
+    onSuccess,
+  });
+
+  return (
+    <TaskCard>
```
jjalan, line 32: This is not expected here. A TaskCard is essentially tied to a task, but here you are showing header, CTA etc. This seems like a poor abstraction.

jjalan:
Can you compare the experience with tailwind admin?

There seems to be different - add task button size is big, the spacing between sidebar and section seems very less and things like that.

https://github.com/jalantechnologies/boilerplate-mern/assets/328858/122ddb7b-67b6-4c42-a933-00c2ed58cde4

jjalan (commented):
Since we are touching common components (ideally should not have), can you ensure login (email, OTP), sign (email, OTP) continues to look Ok and matches the experience in Tailwind Admin?

aatiqJalanSWE:
> Since we are touching common components (ideally should not have), can you ensure login (email, OTP), sign (email, OTP) continues to look Ok and matches the experience in Tailwind Admin?

- Previously, the primary button was primarily designed without its own size property. Upon examining the login/signup feature, it became apparent that it lacked options for different sizes (default, mini, compact, large). Therefore, it is necessary to add these properties so that we can adjust the button size according to our needs.

- In the updated (PR), I have addressed the button consistency throughout the entire codebase, including the login, signup, and forgot password buttons.

src/apps/frontend/pages/tasks/tasks-form.hook.ts
```diff
@@ -0,0 +1,110 @@
...
+  ) => {
+    formik
+      .setFieldValue(fieldName, data)
+      .then()
+      .catch((err) => {
+        onError(err as AsyncError);
+      });
+  };
+
+  const updateFormik = useFormik({
```
KrishnaVAMSI-2003, line 31: better to rename this to `updateTaskFormik`

src/apps/frontend/pages/tasks/index.tsx
```diff
@@ -1,5 +1,59 @@
...
+    deleteTask, getTasks, isGetTasksLoading, setTasksList, tasksList,
+  } = useTaskContext();
+
+  const fetchTasks = async () => {
+    try {
+      await getTasks();
+    } catch (error) {
+      onError(error as AsyncError);
+    }
+  };
```
KrishnaVAMSI-2003, line 26: this method feels redundant, as the error is already handled in the useEffect itself using catch block, can we directly invoke the getTasks method in the useEffect. Also in useEffect there is no need of then, if we are not doing anything on resolve.

src/apps/frontend/pages/tasks/index.tsx
```diff
@@ -1,5 +1,59 @@
...
+
+  useEffect(() => {
+    fetchTasks()
+      .then()
```
KrishnaVAMSI-2003, line 30: can we remove this `then`, as it is not needed, if we are not doing anything on resolve.

src/apps/frontend/pages/tasks/task-header.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  const { addTaskFormik, setFormikFieldValue } = useTaskForm({
+    onError,
+    onSuccess,
+  });
+
+  return (
+    <div className="rounded-sm  border border-stroke bg-white p-3 shadow-default">
+      <div className="flex flex-col items-start justify-between gap-5 sm:flex-row sm:items-center ">
+        <Flex>
```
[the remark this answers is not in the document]
aatiqJalanSWE: In order to make our website responsive, we need to use flex because in mobile mode, our user interface will appear as a column.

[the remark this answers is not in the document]
KrishnaVAMSI-2003: Here it just contains only one item in this flex, so it will act as a simple div. And also the responsiveness is added to its parent, so I think adding the flex to this heading doesn't make any difference.

src/apps/frontend/pages/tasks/task-modal.tsx
```diff
@@ -0,0 +1,127 @@
...
+  task,
+  taskOperationType,
+}) => {
+  const { isAddTaskLoading } = useTaskForm({ onSuccess, onError });
+
+  useEffect(() => {
+    if (taskOperationType === TaskOperationType.EDIT.toString() && task) {
+      setFieldValue(formik, 'title', task.title);
+      setFieldValue(formik, 'id', task.id);
+      setFieldValue(formik, 'description', task.description);
```
KrishnaVAMSI-2003, line 51: we setting this on edit campaign modal right, in that case we can simply set these fields in the method which is toggling the modal(opening the modal) which is `handleTaskOperation` method in `TaskSection` component in this scenario. There will be no need of useEffect if we do this

src/apps/frontend/pages/tasks/task-header.tsx
```diff
@@ -0,0 +1,61 @@
...
+          </Button>
+          <TaskModal
+            formik={addTaskFormik}
+            setFieldValue={setFormikFieldValue}
+            isModalOpen={isModalOpen}
+            setIsModalOpen={setIsModalOpen}
+            btnText={'Add Task'}
+            taskOperationType={TaskOperationType.ADD}
+          />
+        </Flex>
```
KrishnaVAMSI-2003, line 55: Screenshot 2024-03-28 at 11 46 52 AM

I think we can refract and remove the Flex here like this, there is no need to wrap the button and modal with flex too, we can simply place the modal outside and wrap the button with div. Please let me know if this is breaking anything

src/apps/frontend/routes/protected.tsx
```diff
@@ -26,7 +27,14 @@ export const protectedRoutes = [
     element: <App />,
     children: [
       { path: '', element: <Dashboard /> },
```
KrishnaVAMSI-2003, line 29: replace this path with routes constant

KrishnaVAMSI-2003 (approved):

KrishnaVAMSI-2003:
there is no confirmation pop-up/modal while deleting a task, is it not needed?

src/apps/frontend/pages/tasks/task-section.tsx
```diff
@@ -0,0 +1,116 @@
...
+        <Spinner />
+      </div>
+    );
+  }
+
+  return (
+    <VerticalStackLayout gap={7}>
+      {tasks.length > 0 && (
+        <HeadingSmall>
+          To Do's ({tasks.length >= 10 ? Math.floor(tasks.length / 10) : '0'}
```
jjalan, line 64: `{tasks.length >= 10 ? Math.floor(tasks.length / 10) : '0'` - What is this and why is this needed?

src/apps/frontend/services/task.service.ts
```diff
@@ -0,0 +1,73 @@
...
+    } catch (e) {
+      return new ApiResponse(undefined, new ApiError(e.response.data as JsonObject));
+    }
+  };
+
+  deleteTask = async (taskId: string): Promise<ApiResponse<void>> => {
+    const userAccessToken = JSON.parse(
+      localStorage.getItem('access-token'),
+    ) as AccessToken;
+    return this.apiClient.delete(`/tasks/${taskId}`, {
```
jjalan, line 67: Where are you returning a Javascript object of type `ApiResponse`?

src/apps/frontend/services/task.service.ts
```diff
@@ -0,0 +1,73 @@
...
+    } catch (e) {
+      return new ApiResponse(undefined, new ApiError(e.response.data as JsonObject));
+    }
+  };
+
+  getTasks = async (): Promise<ApiResponse<Task[]>> => {
+    const userAccessToken = JSON.parse(
+      localStorage.getItem('access-token'),
+    ) as AccessToken;
+    return this.apiClient.get('/tasks', {
```
jjalan, line 34: Where are you returning type `ApiResponse`? There is no Javascript object being created, rather you are simply return JSON returned from server.

For example, if there is a custom function on `ApiResponse` defined, then the caller of the function will not be able to execute it on the return value type

jjalan (changes requested):

src/apps/frontend/pages/tasks/task-section.tsx
aatiqJalanSWE: image

Logic for the calculation is that tasks are initially counted as 0 at ten's place, then, as tasks increase, the ten's place changes from 0 to 1, and so on.

jjalan (approved):

jjalan:
I am not sure if this is a regression introduced in this PR - but the password field seems to have more height than the normal input field such as name.

Screenshot 2024-04-04 at 9 32 49 AM

jjalan:
Can we check the experience for loading state for a button in Tailwind - https://flowbite.com/docs/components/buttons/?

I would expect, in this case, button to show just spinner, and not `+` and the button s

https://github.com/jalantechnologies/boilerplate-mern/assets/328858/64ee76d9-e0ce-4626-a57b-fd904b069b33

Also, there is no loading indicator while updating a task at all.

hould be disabled

jjalan (approved):
Can you ensure your PR description is accurate before merging?

aatiqJalanSWE:
> I am not sure if this is a regression introduced in this PR - but the password field seems to have more height than the normal input field such as name.
> 
> Screenshot 2024-04-04 at 9 32 49 AM

it appears just because of the secondary button default size padding 
Right now I have removed the initial size of button as earlier is was default in size.

aatiqJalanSWE:
> Can we check the experience for loading state for a button in Tailwind - https://flowbite.com/docs/components/buttons/?
> 
> I would expect, in this case, button to show just spinner, and not `+` and the button s
> 
>  demo.mov 
> Also, there is no loading indicator while updating a task at all.
> 
> hould be disabled

removed the + icon while adding and updating the task.

PR #135, opened 2024-07-23
Feat: Account Deletion Feature

abhishek310001:
## Description
_The goal of this PR is to add the feature of deleting an existing account.
Added page for accessing profile information as well as Accunt related actions._

## Additions
- _Added a Profile Settings page._

## Changes
- Modified `Sidebar Menu Item` to handle Sub Menus.
- Added a common layout to `main` that is used throughout the app

## Screenshots
Screenshot 2024-07-23 at 18 42 45
Screenshot 2024-07-25 at 13 38 40
Screenshot 2024-07-25 at 13 38 47
Screenshot 2024-07-23 at 18 46 41

## Videos

_Video demonstrating usage of Account Deletion feature with responsiveness._

https://github.com/user-attachments/assets/9290d13c-3a59-423c-a145-0826ed911751

## Database schema changes
_NA_

- _DELETE /accounts/:accountId_
  -  should delete an account
  - should throw an error when deleting an account with an account ID that does not exist

### Manual test cases run
- _Checked manually using postman if account is deleting._
- _Checked if account is deleted or not after deleting account from frontend by looking into the db._
- _Since an existing component is modified, checked if didn't break functionality while maintaining the new one._


src/apps/frontend/services/account.service.ts
```diff
@@ -15,4 +15,12 @@ export default class AccountService extends APIService {
       },
     });
   };
+
+  deleteAccount = async (): Promise<ApiResponse<void>> => {
+    const userAccessToken = JSON.parse(
```
bbbneo333, line 20: Think this should be a common logic defined which can be reused in our code. Ideally - This should be within the same code where the token is being saved in storage in the first place.

src/apps/frontend/types/account.ts
```diff
@@ -1,19 +1,26 @@
...
     this.lastName = json.lastName as string;
+    this.phoneNumber = json.phoneNumber as PhoneNumber;
     this.username = json.username as string;
   }
 
   displayName(): string {
     return (`${this.firstName} ${this.lastName}`).trim();
   }
+
+  displayPhoneNumber(): string {
```
bbbneo333, line 23: Think this function should exist in `PhoneNumber`

bbbneo333 (changes requested):
Overall, looks good! Few comments and one feedback on UI - "Delete Account" button should be in red.

src/apps/frontend/types/account.ts
abhishek310001: modified existing PhoneNumber type to get this functionality.

src/apps/frontend/services/account.service.ts
abhishek310001: Made a common logic for accessing, setting and removing access token from storage

abhishek310001:
> Overall, looks good! Few comments and following feedback on UI:
> 
> Screenshot 2024-07-25 at 12 16 02
> 
> * "Delete Account" button can be in red
> * Not sure what's happening - But background color in text field should be same

"Delete Account" is now in red, updated the screenshots.
As for the Input Component I was using default disabled state seems like there was no styling for this state from the custom component causing the issue, edited it to make better.

bbbneo333 (approved):

PR #155, opened 2024-08-07
feat: Add support for default phone number and otp

abhishek310001:
## Description
_This PR is introduced to add support for default phone number along with default OTP so that OTP verifcation can be bypassed when using default phone number._

## Changes
- Added custom env variables for default OTP and default phone number
- `enableSMS` changed to -> `sms.enabled`
- Enabled SMS in preview env config so we can only bypass otp verification using default phone number and otp when using preview link

## Database schema changes
NA

NA

### Manual test cases run
- Verified If default number and OTP works for preview env
- Verified If OTP is displayed in console when in development env
- Able to add default number and otp in development config and able to bypass with default OTP

### DIFF
--- a/config/preview.yml
+++ b/config/preview.yml
@@ -1,9 +1,8 @@
 accounts:
   passwordResetEmailEnabled: true
 
-defaultOTP: '1111'
-
-enableSMS: false
+sms:
+  enabled: false
 
 logger:
   transports: [ 'datadog' ]
--- a/config/preview.yml
+++ b/config/preview.yml
@@ -2,7 +2,7 @@ accounts:
   passwordResetEmailEnabled: true
 
 sms:
-  enabled: false
+  enabled: true
 
 logger:
   transports: [ 'datadog' ]
--- a/config/production.yml
+++ b/config/production.yml
@@ -2,8 +2,9 @@ accounts:
   passwordResetEmailEnabled: true
 
 defaultOTP: '1111'
+defaultPhoneNumber: '9999999999'
 
-enableSMS: false
+enableSMS: true
 
 logger:
   transports: [ 'datadog' ]
--- a/config/production.yml
+++ b/config/production.yml
@@ -1,10 +1,8 @@
 accounts:
   passwordResetEmailEnabled: true
 
-defaultOTP: '1111'
-defaultPhoneNumber: '9999999999'
-
-enableSMS: true
+sms:
+  enabled: true
 
 logger:
   transports: [ 'datadog' ]
--- a/config/testing.yml
+++ b/config/testing.yml
@@ -8,7 +8,7 @@ mailer:
   forgetPasswordMailTemplateId: 'FORGOT_PASSWORD_MAIL_TEMPLATE_ID'
 
 sms:
-  enabled: true
+  enabled: false
 
 mongoDb:
   uri: 'mongodb://localhost:27017/boilerplate-mern-test'
--- a/config/testing.yml
+++ b/config/testing.yml
@@ -7,7 +7,8 @@ mailer:
   defaultEmailName: 'DEFAULT_EMAIL_NAME'
   forgetPasswordMailTemplateId: 'FORGOT_PASSWORD_MAIL_TEMPLATE_ID'
 
-enableSMS: true
+sms:
+  enabled: true
 
 mongoDb:
   uri: 'mongodb://localhost:27017/boilerplate-mern-test'
--- a/src/apps/backend/modules/communication/sms-service.ts
+++ b/src/apps/backend/modules/communication/sms-service.ts
@@ -1,4 +1,5 @@
 import { ConfigService } from '../config';
+import { isDefaultPhoneNumber } from '../util/phone-number-util';
 
 import TwilioService from './internals/twilio-service';
 import { SendSMSParams } from './types';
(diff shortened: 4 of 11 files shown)


src/apps/backend/modules/communication/sms-service.ts
```diff
@@ -7,7 +8,8 @@ export default class SMSService {
   public static async sendSMS(params: SendSMSParams): Promise<void> {
     const enableSMS = ConfigService.getValue('enableSMS');
 
-    if (!enableSMS) {
+    // If SMS is disabled or the recipient phone number is the default phone number, do not send SMS
+    if (!enableSMS || isDefaultPhoneNumber(params.recipientPhone.phoneNumber)) {
```
bbbneo333, line 12: @abhishek310001 - Let's keep this implementation as it was before. The _sms_ can be anything here and it's not tied to OTP. We can have this check on otp/user level instead.

config/production.yml
```diff
@@ -2,8 +2,9 @@ accounts:
   passwordResetEmailEnabled: true
 
 defaultOTP: '1111'
+defaultPhoneNumber: '9999999999'
 
-enableSMS: false
+enableSMS: true
```
bbbneo333, line 7: Can we refactor this config to:

```yaml
sms:
  enabled: true
```

instead?

config/production.yml
```diff
@@ -2,8 +2,9 @@ accounts:
   passwordResetEmailEnabled: true
 
 defaultOTP: '1111'
```
bbbneo333, line 4: We can use `otp` instead for OTP level settings. May be we can do `otp.defaultPhoneNumber`, `otp.defaultOTP` ?

bbbneo333: I have been rethinking the approach. Let's just enable default OTP and PhoneNumber for all instances for now (not just with prod). We can this to default config instead.

src/apps/backend/modules/communication/sms-service.ts
bbbneo333: Also let's add this - If sms is not enabled, show warning using logger - `SMS not enabled. Could not send message - ${params.messageBody}`

src/apps/backend/modules/otp/internal/otp-util.ts
```diff
@@ -21,4 +24,18 @@ export default class OtpUtil {
...
+
+    // If the phone number is the default phone number in production or if we are in a
+    // non-production environment and the default OTP is set in the config, return the default OTP
+    if (
+      (isProdEnv && isDefaultPhoneNumber(phoneNumber))
+      || (!isProdEnv && ConfigService.hasValue('defaultOTP'))
+    ) {
+      return ConfigService.getValue<string>('defaultOTP');
+    }
+    return OtpUtil.generateOtp(OTP_LENGTH);
```
bbbneo333, line 39: For now let's just simplify this to:
- If default OTP and phone number is configured and phone number matches with the config, return default OTP.
- Otherwise generate OTP
- Remove any env based checks

bbbneo333 (changes requested):
Have been rethinking the approach. Let's add default OTP and PhoneNumber for all instances (not just prod). See - https://github.com/jalantechnologies/boilerplate-mern/pull/155#discussion_r1709806982

config/production.yml
abhishek310001: this looks good, modified it!

config/production.yml
abhishek310001: Sure, will update it with this approach.

src/apps/backend/modules/otp/internal/otp-util.ts
abhishek310001: Simplified it.

abhishek310001: There are few things I'm concerned about using this approach:
- In development env either we can only login using `otp.defaultPhoneNumber` and `otp.defaultOTP` set in development config or we need to view OTP from local DB and use it to login.
- The above goes same for the preview env since `sms.enabled` will be set to `false` in both env.

As of Now we can Login/Create account using any Number by entering `otp.defaultOTP` set in config for development env and preview env as well.
Let me know if this works.

src/apps/backend/modules/communication/sms-service.ts
abhishek310001: Added logger warning and reverted to default conditions.

config/testing.yml
```diff
@@ -7,7 +7,8 @@ mailer:
   defaultEmailName: 'DEFAULT_EMAIL_NAME'
   forgetPasswordMailTemplateId: 'FORGOT_PASSWORD_MAIL_TEMPLATE_ID'
 
-enableSMS: true
+sms:
```
bbbneo333, line 10: Should be disabled no?

src/apps/backend/modules/otp/internal/otp-util.ts
bbbneo333: Ideally:
- In development - As the sms is disabled, you should get the OTP from console itself.
- In preview - I have request let's enable SMS there

Will this works?

config/preview.yml
```diff
@@ -1,9 +1,8 @@
 accounts:
   passwordResetEmailEnabled: true
 
-defaultOTP: '1111'
-
-enableSMS: false
+sms:
```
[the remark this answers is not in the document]
abhishek310001: Sure

config/testing.yml
abhishek310001: Yeah, it should be but It was previously enabled so I didn't modify its value.

bbbneo333 (approved):
LGTM!

PR #157, opened 2024-08-08
Share one or more tasks existing users

shravansrinivas:
## Description
_A feature to be able to share one or more tasks with the existing users. To be able to add this functionality, a couple of new endpoints have been created:_

1. POST `/tasks/share-tasks` - To share one or more tasks with users.
> Request Body: `{taskIds: string[], userIds:[]}`
Response: void

2. GET `/accounts` - To fetch all active users to share the task with.

## Database schema changes
_A new field `sharedWith` has been added to the `Task` schema to hold IDs of the users with whom the task is shared_

- _Description of automated test 1_
- _Description of automated test 2_
- _Description of automated test 3_

### Manual test cases run
- _Description of manual test 1_
- _Description of manual test 2_
- _Description of manual test 3_

### DIFF
--- a/src/apps/backend/modules/task/internal/store/task-db.ts
+++ b/src/apps/backend/modules/task/internal/store/task-db.ts
@@ -6,6 +6,7 @@ export interface TaskDB {
   active: boolean;
   description: string;
   title: string;
+  sharedWith: Types.ObjectId[];
 }
 
 export const TaskDbSchema: Schema = new Schema<TaskDB>(
@@ -29,6 +30,12 @@ export const TaskDbSchema: Schema = new Schema<TaskDB>(
       type: String,
       required: true,
     },
+    sharedWith: {
+      type: [Schema.Types.ObjectId],
+      ref: 'Account',
+      default: [],
+      required: true,
+    }
   },
   {
     collection: 'tasks',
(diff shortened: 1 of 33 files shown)


src/apps/backend/modules/task/internal/store/task-db.ts
```diff
@@ -6,6 +6,7 @@ export interface TaskDB {
   active: boolean;
   description: string;
   title: string;
+  sharedWith: Types.ObjectId[];
```
jjalan, line 9: This is not a good practice to store association as an array in the object itself. It makes it hard to query, reason, etc. You want to create a separate collection around this.

From an OOPS standpoint, this also feels somewhat unintuitive. A better way to model this would be via permission and who have access to what resource via permissions.

Ex: `TaskPermission`

This would have permission (READ | WRITE), account_id etc.

src/apps/backend/modules/task/rest-api/task-router.ts
```diff
@@ -13,6 +13,7 @@ export default class TaskRouter extends ApplicationRouter {
     router.post('/', ctrl.createTask);
     router.get('/', ctrl.getTasks);
     router.get('/:id', ctrl.getTask);
+    router.post('/share', ctrl.shareTasks);
```
jjalan, line 16: This is not a CRUD API. You are exposing functional API. `share` is an action, not a verb.

You need to model your operations in terms of entities to leverage OOPS, CRUD etc.

`TaskPermission` is one way (corresponding API would be `POST /accounts/:accountId/tasks/:taskId/permissions` and you pass user id and permission. There are other ways too

PR #169, opened 2024-09-30
Allow automatic module generation for entire backend crud operation using a template

govind-io:
This pull request introduces a new module generation feature and updates the documentation and configuration files to support it. The most important changes include adding a script to generate new modules, updating the `README.md` with the new script, and creating templates for the module structure.

### New Module Generation Feature:

* [`package.json`](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R18): Added a new script `add-module` to generate a new module using `tmb` and included the `@govindgovind852/tmb` package as a dependency. (`[[1]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R18)`, `[[2]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R76)`)
* [`module-template.yaml`](diffhunk://#diff-17afeb5ebc17a88cfbb5f98ff3a53262a1b01c09754b63cbe8415e4a01e10082R1-R43): Added a configuration file for the module generation, defining the structure and templates for the new modules. (`[module-template.yamlR1-R43](diffhunk://#diff-17afeb5ebc17a88cfbb5f98ff3a53262a1b01c09754b63cbe8415e4a01e10082R1-R43)`)

### Documentation Updates:

* [`README.md`](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R45): Added instructions for generating a new module using the `npm run add-module` command. (`[README.mdR45](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R45)`)

### Module Templates:

* [`templates/module`](diffhunk://#diff-85953ccbf1c323c1f0b1e65b3f63f9411a792c19eae4418a33289a910d138273R1-R3): Created various template files for the new module, including `index-template.ts`, `internal/reader-template.ts`, `internal/store/db-template.ts`, `internal/store/repository-template.ts`, `internal/writer-template.ts`, `rest-api/controller-template.ts`, `rest-api/router-template.ts`, `rest-api/serializer-template.ts`, `rest-api/server-template.ts`, `service-template.ts`, and `types-template.ts`. (`[[1]](diffhunk://#diff-85953ccbf1c323c1f0b1e65b3f63f9411a792c19eae4418a33289a910d138273R1-R3)`, `[[2]](diffhunk://#diff-81ee3b093f477f1a4cab05444644c0552e126ca33bda7cd25dc9e7a2d66d9a68R1-R13)`, `[[3]](diffhunk://#diff-b8cceba744c0307ff632488f6ca9c62ecc961031a6365214841a0202d49a1f34R1-R18)`, `[[4]](diffhunk://#diff-c3ba55eb5b4ba170681bfacf05e9596440987fc871351b5619d2f2dc1fa7a2ccR1-R10)`, `[[5]](diffhunk://#diff-bc0cbeaf5d96c3eeee18356715e30d600c51932b6a512cc527df2013c11704dbR1-R5)`, `[[6]](diffhunk://#diff-b00d9fa3651d4ea3d54ad15fa8c17033a0c480243ce3ede525ae07a09048af26R1-R15)`, `[[7]](diffhunk://#diff-1ec56686fac42ec34142b9f80b0226d3e328e0cd2010ef7fdb97401a57a705d2R1-R12)`, `[[8]](diffhunk://#diff-fa206f512dd95624514c67677ecf7fda441aeab379d043ce84358894fce380c7R1-R3)`, `[[9]](diffhunk://#diff-7fdc50d0aee478bb210f6c118d459e9b7a30a7cc2948f797df88fcfdb2ed6c0bR1-R14)`, `[[10]](diffhunk://#diff-415f88999dfa64cddf0483a994185396e2e08d0193a8bc6b52ea829c56642d4aR1-R8)`, `[[11]](diffhunk://#diff-60d397cfbf60e923e758ce0e9ed0ebc0ea19c53ade22bc20cb24671bb1743b32R1-R3)`)

### Minor Fixes:

* [`README.md`](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L109-R110): Fixed formatting for the troubleshooting section. (`[README.mdL109-R110](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L109-R110)`)## Description


govind-io:
@KrishnaVAMSI-2003  Can we please get this PR reviewed? This has been pending for longer than a week now.

KrishnaVAMSI-2003:
> @KrishnaVAMSI-2003 Can we please get this PR reviewed? This has been pending for longer than a week now.

Hey @govind-io sorry missed it and didn't find enough time to review, I will try to review this PR.

@abhishek310001 can you also review this PR please.

govind-io:
> > @KrishnaVAMSI-2003 Can we please get this PR reviewed? This has been pending for longer than a week now.
> 
> Hey @govind-io sorry missed it and didn't find enough time to review, I will try to review this PR.
> 
> Could you please add a video of manual testing please, when you find some time. Also I have noticed some schemas and code related stating `Test`, `TestDB`, etc., do we need that as well?
> 
> @abhishek310001 can you also review this PR please.

@KrishnaVAMSI-2003  @abhishek310001  the review is still pending, can we review it soon so that the PR does not becomes stale and the team can start using this and benefit from it?

templates/module/service-template.ts
```diff
@@ -0,0 +1,37 @@
+import $entityNameReader from './internals/$moduleName-reader';
+import $entityNameWriter from './internals/$moduleName-writer';
+
+import { $entityName, Get$entityNameParams, PaginatedResponse } from './types';
+
+export default class $moduleNameService {
+  public static get$entityNames({
+    page,
+    size,
```
KrishnaVAMSI-2003, line 9: Can we de-structure these params inside the method itself.

templates/module/service-template.ts
```diff
@@ -0,0 +1,37 @@
...
+  }: Get$entityNameParams): Promise<PaginatedResponse<$entityName>> {
+    return $entityNameReader.get$entityNames({
+      page: page || 1,
+      size: size || $pageLimit,
```
KrishnaVAMSI-2003, line 13: Can we pass the default values from the controller itself. Also can we rename `$pageLimit` to `$defaultPageSize` or similar.

templates/module/service-template.ts
```diff
@@ -0,0 +1,37 @@
...
+    });
+  }
+
+  public static get$entityNameById(id: string): Promise<$entityName> {
```
KrishnaVAMSI-2003, line 17: Can we create the type-def for the params of this and below methods too similar to the above `get$entityNames` method.

templates/module/rest-api/router-template.ts
```diff
@@ -0,0 +1,24 @@
...
+export default class $entityNameRouter extends ApplicationRouter {
+  configure(): void {
+    const { router } = this;
+    const ctrl = new $entityNameController();
+
+    router.use(accessAuthMiddleware);
+
+    const baseUrl='$baseUrl'
+
+    // Define get routes
```
KrishnaVAMSI-2003, line 15: Can we remove these comments.

templates/module/rest-api/router-template.ts
```diff
@@ -0,0 +1,24 @@
...
+import { $entityNameController } from './$moduleName-controller';
+
+export default class $entityNameRouter extends ApplicationRouter {
+  configure(): void {
+    const { router } = this;
+    const ctrl = new $entityNameController();
+
+    router.use(accessAuthMiddleware);
+
+    const baseUrl='$baseUrl'
```
KrishnaVAMSI-2003, line 13: I don't think this is needed as the base url is usually be similar to the module name(similar to the existing server url). But if you think this has some use-case, can we mover this to server class. As, it seems like the baseUrl is added to all the routes.

templates/module/rest-api/serializer-template.ts
```diff
@@ -0,0 +1,3 @@
+import { $entityName } from '../types';
+
+export const serialize$entityNameAsJSON = (data: $entityName): unknown => data;
```
KrishnaVAMSI-2003, line 3: Can we make this serialize method similar to the one's in boilerplate if possible, ex.
```
export const serializeAccountAsJSON = (account: Account): unknown => ({
  id: account.id,
  phoneNumber: account.phoneNumber,
  firstName: account.firstName,
  lastName: account.lastName,
  username: account.username,
});
```
Because in here we are just changing the type of the object, not exactly serializing it.

KrishnaVAMSI-2003 (commented):
Left some comments, along with these
- Could you please make the PR description a bit more readable.

Also there are some earlier questions yet to be addressed
- Could you please add a video of manual testing, when you find some time.
- Also there are some schemas and type defs: Test, TestDB, etc., do we need that as well? I assume those are being used as the default schema/type def for the module and we cannot define the type def, am I correct? If yes can we try to update the implementation to add support for defining schema too.

templates/module/dbSchema.ts
jjalan, line 1: Nit: file name should be case insensitive.

templates/module/dbSchema.ts
```diff
@@ -0,0 +1,14 @@
+{
```
jjalan, line 1: Nit: use 2 spaces instead of 4 spaces.

templates/module/rest-api/serializer-template.ts
govind-io: To serialise it exactly we also need to be aware of the individual property names, but we don't know exactly which field the user wants to send and which he doesn't so here we are just changing the type and letting the user decide which fields the user wants to serialise and send and which don't

jjalan:
@KrishnaVAMSI-2003  - Can you PTAL at your comments and in general the PR?

templates/module/rest-api/serializer-template.ts
KrishnaVAMSI-2003: > To serialise it exactly we also need to be aware of the individual property names

I think this is the same case with defining the type def of the schema too, where we need to know the attribute names. So, we can use similar implementation IMO.

> but we don't know exactly which field the user wants to send and which he doesn't so here we are just changing the type

Usually we were including all the attributes while serialising. In case, where the object's structure matches the desired JSON format exactly and it doesn't contain any sensitive information, we can use the serialise method that you have added. But in contradiction to your statement: `letting the user decide which fields the user wants to serialise and send and which don't` this takes out the control as we cannot choose the fields that we want, and we need to again implement the way that I have stated in the earlier comment. IMO, although it is simple and works well there is also chances that this might
- Exposes the internal structure of the object, which might include unnecessary or sensitive information. 
- Doesn't provide fine-grained control over the serialised output.
- It is also completely ignoring the nested object serialising, ex. let's say there is a resource object inside an account, then this isn't properly serialising the resource object.

So, at the end we might need to implement the template of existing boilerplate's serialiser methods, let's try to align with the same everywhere to keep it consistent. To keep it simple we can just include all the attributes while serialising (I'm fine with this for now, as we can update later with minor efforts), but there are still possibilities of issues 1 and 3. We need to implement some kind of solution like defining the required properties in `yml` file or similar.

PR #173, opened 2024-10-25
Add an api & implementation to generate documentation using ai & seed it under assets. And add a route to view generated api documentation.

KrishnaVAMSI-2003:
## Description
This PR introduces a route `/api/documentation` on which the documentation for all the server api's can be found. On starting the server documentation will be generated and gets injected under `/dist/assets/documentation` folder. And the frontend will use that file to display the generated documentation.

And a server end point `/api/get-documentation` at which the server generates the api documentation using openai and sends in response.

## Database schema changes
- NA

- should return the documentation when documentation is enabled for the environment
- should return an empty string, as the documentation is disabled for the testing environment
- should throw an error when it is unable to access the required files

### Manual test cases run
- Navigate to `{host}/api/documentation` route to access the api documentation of the application. Manual testing [video](https://jalantechnology-my.sharepoint.com/:v:/g/personal/krishna_vamsi_bettrsw_com/ETqvwYovH3RGv3uN0nrmVrYBzfzbDqwOybH_GKo-TRsMuQ?e=eobdGR&nav=eyJyZWZlcnJhbEluZm8iOnsicmVmZXJyYWxBcHAiOiJTdHJlYW1XZWJBcHAiLCJyZWZlcnJhbFZpZXciOiJTaGFyZURpYWxvZy1MaW5rIiwicmVmZXJyYWxBcHBQbGF0Zm9ybSI6IldlYiIsInJlZmVycmFsTW9kZSI6InZpZXcifX0%3D).


src/apps/backend/app.ts
```diff
@@ -42,6 +42,21 @@ export default class App {
     return Promise.resolve(server);
   }
 
+  public static getAllServers(): (
+    | AccountServer
```
jjalan, line 46: This can quickly get out of hand, can the return type be array of ApplicationServer?

src/apps/backend/documentation/express-list-routes.ts
jjalan, line 1: Most probably should move to `internal` folder

src/apps/backend/documentation/index.ts
```diff
@@ -0,0 +1,13 @@
+import App from '../app';
+
+import expressListRoutes from './express-list-routes';
+
+(() => {
+  const servers = App.getAllServers();
```
jjalan, line 6: You want to still DocumentationService and do the logic there and have it call here.

src/apps/backend/modules/access-token/rest-api/access-token-server.ts
```diff
@@ -3,6 +3,10 @@ import { ApplicationServer } from '../../application';
 import AccessTokenRouter from './access-token-router';
 
 export default class AccessTokenServer extends ApplicationServer {
+  constructor() {
+    super(new AccessTokenRouter().routerFilePath);
```
jjalan, line 7: We want to avoid this

src/apps/backend/app.ts
jjalan: If the path thing keeps failing, you can update the implementation here 

`public static getAPIMicroservices(): APIMicroserviceService[] {}`

Where:

```
type APIMicroserviceService {
  serviceName: string;
  serviceRootFolderName: string;
  serverInstance: ApplicationServer
}
```

src/apps/backend/app.ts
```diff
@@ -11,18 +11,21 @@ import { ConfigService } from './modules/config';
 import { Logger, CustomLoggerTransport } from './modules/logger';
 import { PasswordResetTokenServer } from './modules/password-reset-token';
 import { TaskServer } from './modules/task';
+import { APIMicroserviceService } from './types';
 
 const isDevEnv = process.env.NODE_ENV === 'development';
 
 export default class App {
+  public static baseRoutePath = '/api';
```
jjalan, line 19: rename to this `baseAPIRoutePath`

src/apps/backend/documentation/types.ts
jjalan, line 1: This file should contain def that this file exposes to other module.

None of them seem like they should be exposed I think, may consider moving this file to internals folder.

package.json
```diff
@@ -7,11 +7,12 @@
     "npm": "10.2.4"
   },
   "scripts": {
-    "build": "cross-env NODE_ENV=production concurrently --kill-others-on-fail npm:build:*",
+    "build": "npm run generate-documentation && cross-env NODE_ENV=production concurrently --kill-others-on-fail npm:build:*",
```
jjalan, line 10: Is this called only from CI during release?

I want to make sure we are mindful of calling this because it will eventually call into OpenAI, and may cost money.

src/apps/backend/documentation/documentation-service.ts
```diff
@@ -0,0 +1,36 @@
...
+    );
+    let documentationContent = '';
+
+    apiMicroservices.forEach((server) => {
+      const routes = expressListRoutes(server.serverInstance.server);
+      const serverInfo = `Server Root: ${server.serverRootFolderPath}\nRoutes:\n${JSON.stringify(routes, null, 2)}\n`;
+      documentationContent += serverInfo;
+    });
+
+    const isDocumentationEnabled = ConfigService.getValue<boolean>(
```
jjalan, line 26: Probably add a cheeck for this at the beginning of this method and bail if not enabled.

src/apps/backend/documentation/documentation-service.ts
```diff
@@ -0,0 +1,297 @@
...
+      'documentation.enabled',
+    );
+
+    if (isDocumentationEnabled) {
+      const documentationPath = path.join(
+        __dirname,
+        '../../../assets/documentation/index.md',
+      );
+      const routes = this.getAllApiRoutesWithDetails();
+      const prompt = `Generate a comprehensive API documentation in a human readable format, for the following routes: ${JSON.stringify(routes, null, 2)}. And return the documentation as a markdown file, without any additional text or comments like adding \`\`\`markdown at the beginning or at the end. Also please provide the response object in simplified JSON format, without any typescript code or comments.`;
```
jjalan, line 24: Let's move the prompt into a constant so its easy to change.

src/apps/backend/documentation/internals/internal-types.ts
jjalan, line 1: Name the file as `types.ts`. Its already inside internals folder, so `internal-` in the file name is moot.

src/apps/backend/documentation/rest-api/openai-adapter.ts
jjalan, line 1: Let's create an OpenAI module, service and adapter.

routes-sample-log.json
jjalan, line 1: Does this file probably needs to be removed and/or gitignored?

package.json
KrishnaVAMSI-2003: > Is this called only from CI during release?

We can configure that from the config files sir. Once done with proper testing I will disable this for preview env.

routes-sample-log.json
KrishnaVAMSI-2003: Yes sir, this file should be removed, I just placed it here, so that you can verify the data that we are sending to OpenAI. I will remove this.

src/apps/backend/app.ts
```diff
@@ -42,6 +45,40 @@ export default class App {
     return Promise.resolve(server);
   }
 
+  public static getAPIMicroservices(): APIMicroserviceService[] {
+    const microservices: APIMicroserviceService[] = [];
+
+    // add your new server here to the list
+    const servers = [
```
jjalan, line 52: Can we move this in a separate function?

I think for all practical purposes, during application development, I expect this to change, and not rest of the code in this function.

src/apps/backend/modules/openai/openai-service.ts
```diff
@@ -0,0 +1,10 @@
+import OpenAIAdapter from './rest-api/openai-adapter';
+import { Nullable } from './types';
+
+export default class OpenAIService {
+  public static async generateDocumentation(
```
jjalan, line 5: Is it possible to make this method a bit more generic that accepts like a prompt and return a response?

src/apps/backend/types.ts
jjalan, line 1: Can we move this inside application or other common module?

src/apps/frontend/webpack.dev.js
```diff
@@ -22,7 +22,14 @@ const config = {
     port: devServerPort,
     proxy: {
       secure: false,
-      '/api': `http://localhost:${devServerAPIProxyPort}`,
+      '/api': {
```
jjalan, line 25: Can you add please share why this change is needed? My knowledge of webpack is bit dated, rusty at the moment.

src/apps/frontend/webpack.base.js
```diff
@@ -65,6 +65,10 @@ module.exports = {
           },
         ],
       },
+      {
+        test: /\.md$/,
```
jjalan, line 69: Can you add please share why this change is needed? My knowledge of webpack is bit dated, rusty at the moment.

src/apps/backend/modules/documentation/documentation-service.ts
```diff
@@ -0,0 +1,271 @@
...
+
+    const routes = this.getRoutesWithControllerAndSerializerDetails();
+    const prompt = `${DOCUMENTATION_GENERATION_PROMPT}\n${JSON.stringify(routes, null, 2)}`;
+    return {
+      markdownDocumentation:
+        await OpenAIService.getChatCompletionResponse(prompt),
+    };
+  }
+
+  private static getRoutesWithControllerAndSerializerDetails(): HttpRouteWithControllerAndSerializerDetails[] {
```
jjalan, line 42: In an ideal world, it would be good to move these into a new file - `DocumentGenerator.ts` inside internals and have the service call into it.

This is causing the service file to bloat and exposes a lot more internal implementation details when someone looks at the file.

src/apps/frontend/webpack.dev.js
KrishnaVAMSI-2003: This `proxy` is being used to forward the requests to the server which contains `/api` in their url. To place the documentation page on this route, we should bypass this specific route. I have added the comments accordingly sir.

src/apps/frontend/webpack.base.js
KrishnaVAMSI-2003: This is added to load the readme files in earlier implementation, I have removed it now.

src/apps/backend/modules/documentation/documentation-service.ts
KrishnaVAMSI-2003: Moved this and respective private methods to `document-generator-util.ts` file sir.

src/apps/backend/app.ts
```diff
@@ -52,6 +52,8 @@ export default class App {
       Logger.info('app - server started listening on  - %s', server.address());
     });
 
+    await DocumentationService.generateAndInjectDocumentationUnderDistFolder();
```
jjalan, line 55: Can we not await and let it happen async so server can get started?

src/apps/backend/app.ts
```diff
@@ -39,9 +52,43 @@ export default class App {
       Logger.info('app - server started listening on  - %s', server.address());
     });
 
+    DocumentationService.generateAndInjectDocumentationUnderDistFolder().catch(
```
jjalan, line 55: rename `generateAndInjectDocumentationUnderDistFolder` to `generateAPIDocumentation`. Dist folder is somewhat an internal concern of the service.

PR #198, opened 2025-02-03
fix: Enable strictNullChecks, noImplicitAny checks in TS config

### LINKED ISSUE #139: fix: Enable strictNullChecks, noImplicitAny checks in TS config
Currently we have following in our TS config:

```json
"strictNullChecks": false,
"noImplicitAny": false
```

This can lead to following issues:
- More runtime errors due to unchecked null or undefined values.
- Less type safety and more bugs, as the compiler can't catch implicit any types or null issues.
- Harder code maintenance and debugging due to less clear type expectations.

We should enable them to have strict type policy.

sn1f3rt:
## Description
Resolves #139 

This pull request enhances the TypeScript configuration by enabling the `strictNullChecks` and `noImplicitAny` compiler options. These changes aim to improve type safety and code quality across the codebase.​

## Database schema changes

N/A

### Manual test cases run

- Manually tested all the existing functionalities on both localhost and Preview App.
- Confirmed no runtime errors or console warnings related to null/undefined values after changes.

### DIFF
--- a/test/spec/task/task.spec.ts
+++ b/test/spec/task/task.spec.ts
@@ -195,7 +195,6 @@ describe('Task API', () => {
 
       const updatedToken = await TaskRepository.findById(task.id);
 
-      expect(updatedToken).to.exist;
       expect(updatedToken!.active).to.be.false;
     });
   });


.editorconfig
```diff
@@ -2,7 +2,7 @@ root = true
 
 [*]
 charset = utf-8
-end_of_line = lf
+end_of_line = crlf
```
sn1f3rt, line 5: In case anyone is wondering why this change is required, the editorconfig mentions LF line endings; however, all the files in this repo have CRLF line endings, which in essence means when the formatter is correctly run after the lint staged PR got merged, it adds an additional 200-ish files changes which are solely CRLF to LF conversions.  

I will revert this to LF (which it should be), and add the format changes in a later commit once this is approved, otherwise it will be a nightmare to review the files when 250 files show up with majority remaining unchanged except a line ending change.

jjalan:
For frontend changes, can you confirm it is *exactly* same as the changes [here](https://github.com/jalantechnologies/rflask-boilerplate/pull/83)?

cypress/e2e/inspectlet.spec.cy.ts
```diff
@@ -17,7 +17,7 @@ xdescribe('Inspectlet', () => {
   it('should disable integration if key was not provided', () => {
     cy.visit('/?inspectlet_diagnostics=true', {
       onLoad(window) {
-        set(window, 'Config.inspectletKey', '');
+        set(window, 'Config.inspectLetKey', '');
```
jjalan, line 20: This seems unrelated change to this PR

cypress/e2e/login.spec.cy.ts
```diff
@@ -15,8 +16,10 @@ describe('Login', () => {
 
   it('should allow login', () => {
     cy.get('[data-testid="username"]').clear();
+    // @ts-expect-error - credentials might be null and supposed to be that way for the cypress test
```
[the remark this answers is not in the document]
jjalan: Use TS null assertion operator perhaps?

src/apps/backend/modules/access-token/access-token-service.ts
```diff
@@ -63,7 +63,11 @@ export default class AccessTokenService {
       throw new AccessTokenInvalidError();
     }
 
-    if (verifiedToken.exp * 1000 < Date.now()) {
+    if (
+      !verifiedToken ||
```
jjalan, line 67: Is `verifiedToken` can be nil if the code reaches here? If not, look into use TS assertion operator instead of adding more branching conditions in the code using `||`

sn1f3rt:
> For frontend changes, can you confirm it is _exactly_ same as the changes [here](https://github.com/jalantechnologies/rflask-boilerplate/pull/83)?

Synced via commit https://github.com/jalantechnologies/boilerplate-mern/pull/198/commits/3bacbf025dbfe78790fd34a584fe5e29ca215663

src/apps/backend/modules/access-token/access-token-service.ts
sn1f3rt: Yeah, that check is redundant. Removed via https://github.com/jalantechnologies/boilerplate-mern/pull/198/commits/df39e253b15894ab12ab167be4a20d427a9d11b2

cypress/e2e/inspectlet.spec.cy.ts
sn1f3rt: It was being reported as being wrongly cased, so I included it too. Should I remove it?

cypress/e2e/login.spec.cy.ts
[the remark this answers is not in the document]
sn1f3rt: Resolved https://github.com/jalantechnologies/boilerplate-mern/pull/198/commits/15a090c6708ef466831ca70962afa3299984053b

src/apps/backend/modules/access-token/access-token-service.ts
```diff
@@ -88,7 +88,11 @@ export default class AccessTokenService {
 
     const jwtTokenDecoded = jsonwebtoken.decode(jwtToken);
 
-    if (typeof jwtTokenDecoded === 'string') {
+    if (
```
jjalan, line 91: Is this change related strictly to the scope of this PR or is this surfacing an existing bug in the system. If its latter, it should be addressed in a separate PR

cypress/e2e/login.spec.cy.ts
```diff
@@ -1,8 +1,9 @@
 import { CreateAccountParamsByUsernameAndPassword } from '../../src/apps/backend/modules/account';
 import { setupScenario } from '../helpers/scenario';
+import { Nullable } from '../types/common';
 
 describe('Login', () => {
-  const credentials: CreateAccountParamsByUsernameAndPassword =
+  const credentials: Nullable<CreateAccountParamsByUsernameAndPassword> =
```
jjalan, line 6: It is not clear why, or more importantly, when it should be `Nullable`

cypress/helpers/scenario.ts
```diff
@@ -1,8 +1,10 @@
+import { Nullable } from '../types/common';
+
 export const setupScenario = <T = unknown>(
   name: string,
   params?: unknown,
-): T => {
-  let result: T;
+): Nullable<T> => {
```
jjalan, line 6: It is not clear why, or more importantly, when it should be `Nullable`

src/apps/backend/modules/access-token/access-token-service.ts
```diff
@@ -63,7 +63,7 @@ export default class AccessTokenService {
       throw new AccessTokenInvalidError();
     }
 
-    if (verifiedToken.exp * 1000 < Date.now()) {
+    if (!verifiedToken.exp || verifiedToken.exp * 1000 < Date.now()) {
```
jjalan, line 66: Why or when `verifiedToken.exp` is nil? I do not expect it not to be set, if there is an instance of token.

jjalan (commented):
I have left some cursory comments and did not do a complete PR review. Based on what I saw, I think it is important to consider if something can really be null or undefined before adding bunch of if checks.

src/apps/backend/modules/access-token/access-token-service.ts
sn1f3rt: That is because in the dependency module typing for jwtpayload, exp is set as optional, so typescript flags that it may possibly be undefined
image

src/apps/backend/modules/access-token/access-token-service.ts
sn1f3rt: Yes, same reason as https://github.com/jalantechnologies/boilerplate-mern/pull/198/files/15a090c6708ef466831ca70962afa3299984053b#r1946478991

src/apps/backend/modules/access-token/access-token-service.ts
jjalan: Right.

However, since we do not expect it to be nil by the time the code reaches here, should it be?

`if (!verifiedToken.exp * 1000 < Date.now()) {`

cypress/e2e/login.spec.cy.ts
sn1f3rt: That would be the case in the cleanup scenario https://github.com/jalantechnologies/boilerplate-mern/blob/sayan/fix/strict-typing/cypress/helpers/scenario.ts#L18

cypress/helpers/scenario.ts
sn1f3rt: That would be the case in the cleanup scenario https://github.com/jalantechnologies/boilerplate-mern/blob/sayan/fix/strict-typing/cypress/helpers/scenario.ts#L18

cypress/e2e/login.spec.cy.ts
```diff
@@ -15,26 +16,36 @@ describe('Login', () => {
 
   it('should allow login', () => {
     cy.get('[data-testid="username"]').clear();
-    cy.get('[data-testid="username"]').type(credentials.username);
-    cy.get('[data-testid="password"]').clear();
-    cy.get('[data-testid="password"]').type(credentials.password);
-    cy.get('button[data-baseweb="button"]').click();
 
-    cy.url().should('be.equal', `${Cypress.config('baseUrl')}/`);
+    if (credentials) {
```
jjalan, line 20: If should not be needed, when I read the code above, it seems like credentials are only null when the specs are ended.

Adding if condition adds additional branching when code runs which is not desireable.

We should be able to use non null assertion operation of TS `!` without increasing the branching condition. We just need to tell TS that we know what we are doing and credentials should not be null when its being accessed here.

sn1f3rt: Updated via https://github.com/jalantechnologies/boilerplate-mern/pull/198/commits/f32d8c42f28f9fca8b414dea6dc54193f49ce83c

src/apps/backend/modules/access-token/rest-api/access-token-serializer.ts
```diff
@@ -1,9 +1,10 @@
+import { Nullable } from '../../../types';
 import { AccessToken } from '../types';
 
 export const serializeAccessTokenAsJSON = (
-  accessToken: AccessToken,
+  accessToken: Nullable<AccessToken>,
```
jjalan, line 5: This does not sound right.

By the time code calls this function, `accessToken` should not be nullable

jjalan (commented):
I haven't reviewed the PR in full, but I think we need to audit these changes with the point of view that when something should be nullable by reading the code and use TS assertion operator.

src/apps/backend/modules/access-token/access-token-service.ts
sn1f3rt: Resolved via 9d4e92c183583e4a4634e900877be4cb1e86c864

src/apps/backend/modules/access-token/rest-api/access-token-serializer.ts
sn1f3rt: Resolved via 9d4e92c183583e4a4634e900877be4cb1e86c864

src/apps/backend/modules/access-token/rest-api/access-token-serializer.ts
```diff
@@ -1,9 +1,10 @@
...
 ): unknown => ({
-  accountId: accessToken.accountId,
-  expiresAt: accessToken.expiresAt.toUTCString(),
-  token: accessToken.token,
+  accountId: accessToken?.accountId,
```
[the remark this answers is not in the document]
sn1f3rt: Yes sure, I took the liberty to review my code and found that this could be applied to a couple other places as well, I've pushed the updates via 9d4e92c183583e4a4634e900877be4cb1e86c864

src/apps/backend/modules/account/internal/account-writer.ts
```diff
@@ -72,6 +72,10 @@ export default class AccountWriter {
       { new: true },
     );
 
+    if (!dbAccount) {
```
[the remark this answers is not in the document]
sn1f3rt: Right, the check is already being done elsewhere, resolved via 9d4e92c183583e4a4634e900877be4cb1e86c864

src/apps/backend/modules/account/rest-api/account-serializer.ts
```diff
@@ -1,9 +1,12 @@
...
 import { Account } from '../types';
 
-export const serializeAccountAsJSON = (account: Account): unknown => ({
-  id: account.id,
-  phoneNumber: account.phoneNumber,
-  firstName: account.firstName,
-  lastName: account.lastName,
-  username: account.username,
+export const serializeAccountAsJSON = (
+  account?: Nullable<Account>,
```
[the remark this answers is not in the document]
sn1f3rt: Resolved via 9d4e92c183583e4a4634e900877be4cb1e86c864

src/apps/backend/modules/account/internal/account-writer.ts
```diff
@@ -64,13 +65,14 @@ export default class AccountWriter {
...
-    const dbAccount = await AccountRepository.findByIdAndUpdate(
+    // Type assertion for AccountDB because accountId is already validated to be present in the db
+    const dbAccount = (await AccountRepository.findByIdAndUpdate(
       accountId,
       {
         hashedPassword: accountHashedPassword,
       },
       { new: true },
-    );
+    )) as AccountDB;
```
jjalan, line 75: Is there a way to use generics, override  `findByIdAndUpdate ` by `AccountRepository` such that it automatically returns the right type instead of every consumer uses `as`?

govind-io: @jjalan, ideally, we should add an `if` statement to check whether `dbAccount` is `undefined`. While the parent `function` is already validating it, the write `function` should not rely on external validation. Instead, it should independently ensure that all necessary conditions are met before proceeding.

By adding this check, we can eliminate the need for a TypeScript assertion. TypeScript currently warns that `dbAccount` could be undefined, and we're bypassing this warning with an assertion, assuming the parent `function` has validated it. However, since we have no control over the parent's validation, it's best to handle the check within the writer `function` itself.

jjalan: > it's best to handle the check within the writer function itself.

I see, in that case, it sounds like we should throw if `dbAccount` is not set

govind-io: Yeah, that's correct inside the `if` check we should throw an error.

sn1f3rt: Update via https://github.com/jalantechnologies/boilerplate-mern/pull/198/commits/c70e898aac7eb60557c12ffb3d7edfd703fe9e77

src/apps/backend/modules/documentation/internals/document-generator-util.ts
```diff
@@ -159,8 +162,10 @@ export default class DocumentGeneratorUtil {
   }
 
   private static addLeadingSlashesIfNotExistsAndRemoveTrailingSlashes(
-    routePath: string,
-  ): string {
+    routePath: Nullable<string>,
+  ): Nullable<string> {
+    if (!routePath) return null;
+
```
govind-io, line 168: The previous type def here seems to be non nullable, why are we changing the nature of existing type?  Seems like the `routePath` indeed should be just `string`

src/apps/backend/modules/account/internal/account-writer.ts
```diff
@@ -72,6 +72,10 @@ export default class AccountWriter {
       { new: true }
     );
 
+    if (!dbAccount) {
+      throw new AccountNotFoundError(`Account with id ${accountId} not found`);
```
jjalan, line 76: The error message here and other places is repetitive. Can we pass accountId in the constructor and construct a better error message in `AccountNotFoundError ` itself - ex: `System is unable to find an account with id: {accountId}. Please check it and specify a valid id to continue`

src/apps/backend/modules/documentation/internals/document-generator-util.ts
```diff
@@ -75,9 +76,9 @@ export default class DocumentGeneratorUtil {
   }
 
   private static getSerializerMethodCode(
-    controllerMethodCode: string,
+    controllerMethodCode: Nullable<string>,
```
jjalan, line 79: I dont think this should be nullable. We probably want to use `!` operator at the caller site. Same for `extractSerializeMethodName `. Hence we should revert the changes in this file completely.

src/apps/backend/modules/documentation/types.ts
```diff
@@ -8,10 +9,10 @@ export type HttpRouteWithRootFolderPath = {
 };
 
 export type HttpRouteWithControllerAndSerializerDetails = {
```
jjalan, line 11: Revert changes in this file.

src/apps/backend/modules/password-reset-token/internal/password-reset-token-writer.ts
```diff
@@ -26,12 +26,22 @@ export default class PasswordResetTokenWriter {
...
+    const updatedToken = await PasswordResetTokenRepository.findByIdAndUpdate(
       passwordResetTokenId,
       {
         isUsed: true,
       },
       { new: true }
     );
+
+    if (!updatedToken) {
+      throw new PasswordResetTokenNotFoundError(
```
jjalan, line 38: Move the error message inside the `PasswordResetTokenNotFoundError ` and have `PasswordResetTokenNotFoundError ` accept `passwordResetTokenId`.

Error message should be: `System is unable to find the reset token with id: {passwordResetTokenId}. Kindly recheck it and try again`

src/apps/frontend/components/typography/heading-large.tsx
```diff
@@ -1,7 +1,11 @@
-import React, { PropsWithChildren } from 'react';
+import React, { ReactNode, PropsWithChildren } from 'react';
 
-const HeadingLarge: React.FC<PropsWithChildren> = ({ children }) => (
-  <h2 className="text-[28px] font-semibold text-black">{children}</h2>
-);
+interface HeadingLargeProps {
+  children: ReactNode;
```
jjalan, line 4: You do not need this.

`PropsWithChildren` adds this automatically. 

This is same for other places in this PR where `children: ReactNode;` is explicitly defined.

src/apps/backend/modules/account/internal/account-writer.ts
sn1f3rt: Updated via 3000dd052747cc1d950026d892562f218bb5f5c7

src/apps/backend/modules/password-reset-token/internal/password-reset-token-writer.ts
sn1f3rt: Updated via 7c8315b3c40b67187d2579719c21d8bf576c29b3

src/apps/backend/modules/documentation/internals/document-generator-util.ts
sn1f3rt: We cannot revert the entire file. If you check line 145 (the last line on `extractSerializeMethodName`), it can possibly return null. I've changed rest of the null assertions via https://github.com/jalantechnologies/boilerplate-mern/pull/198/commits/636fc44e0dcacb2591d231978138ff56e2a20403

src/apps/backend/modules/documentation/types.ts
sn1f3rt: Like I said above, the second Nullable cannot re reverted due to the extract method constraint.

src/apps/frontend/components/typography/heading-large.tsx
saisudan2003: Updated via [1942d5a](https://github.com/jalantechnologies/boilerplate-mern/pull/198/commits/1942d5a58c55678a01086acea006572dc4afc06e)

src/apps/frontend/pages/tasks/tasks-form.hook.ts
```diff
@@ -62,7 +62,7 @@ const useTaskForm = ({ onError, onSuccess }: TaskFormProps) => {
             }
             return taskData;
           });
-          setTasksList(newUpdatedTasks);
+          setTasksList(newUpdatedTasks as Task[]);
```
jjalan, line 65: In an ideal world, `as` should not be required. This tells me somewhere the typedef is not declared correctly due to which TS is unable to determine that `newUpdatedTasks ` is a list of `Task`. Can we please double check this code path? `as` is effectively we forcefully telling TS the type which is not desireable.

src/apps/frontend/pages/tasks/tasks-form.hook.ts
```diff
@@ -90,7 +90,7 @@ const useTaskForm = ({ onError, onSuccess }: TaskFormProps) => {
     onSubmit: (values) => {
       addTask(values.title, values.description)
         .then((newTask) => {
-          setTasksList([...tasksList, newTask]);
+          setTasksList([...tasksList, newTask as Task]);
```
jjalan, line 93: In an ideal world, `as` should not be required. This tells me somewhere the typedef is not declared correctly due to which TS is unable to determine that `newTask ` is a`Task`. Can we please double check this code path? `as` is effectively we forcefully telling TS the type which is not desirable.

src/apps/frontend/pages/authentication/phone-login/phone-login-form.hook.ts
```diff
@@ -39,21 +39,20 @@ const usePhoneLoginForm = ({
...
-        .getNationalNumber()
-        .toString();
       const encodedCountryCode = encodeURIComponent(values.countryCode);
       const otpPageUrl = `${routes.VERIFY_OTP}?&country_code=${encodedCountryCode}&phone_number=${formattedPhoneNumber}`;
 
       sendOTP(
         new PhoneNumber({
           countryCode: values.countryCode,
-          phoneNumber: formattedPhoneNumber,
+          phoneNumber: formattedPhoneNumber.toString(),
```
jjalan, line 55: Is this a bug? Seems more than a TS fix.

src/apps/backend/modules/documentation/internals/document-generator-util.ts
```diff
@@ -136,7 +137,9 @@ export default class DocumentGeneratorUtil {
 
   private static extractSerializeMethodName(
     controllerMethodCode: string
-  ): string | null {
+  ): Nullable<string> {
+    if (!controllerMethodCode) return null;
```
jjalan, line 141: This does not look right.

`controllerMethodCode` is type of string (and cannot be null or undefined). Why is this required?

src/apps/frontend/pages/tasks/tasks-form.hook.ts
saisudan2003: Updated via commit [4dc3f41](https://github.com/jalantechnologies/boilerplate-mern/pull/198/commits/4dc3f415a389a07c7f8fb8b7723b8c2bfc0b900d)

`updateTask` returns a `Nullable<Task>,` I’ve added a null check before calling `setTasksList`. Now we only proceed when the response is non-null, removing the need for the forced `as Task[]` cast.

src/apps/frontend/pages/tasks/tasks-form.hook.ts
saisudan2003: Updated via commit [4dc3f41](https://github.com/jalantechnologies/boilerplate-mern/pull/198/commits/4dc3f415a389a07c7f8fb8b7723b8c2bfc0b900d)

`addTask` returns a `Nullable<Task>,` I’ve added a null check before calling `setTasksList`. Now we only proceed when the response is non-null, removing the need for the forced `as Task` cast.

src/apps/backend/modules/documentation/internals/document-generator-util.ts
saisudan2003: Updated via commit [4dc3f41](https://github.com/jalantechnologies/boilerplate-mern/pull/198/commits/4dc3f415a389a07c7f8fb8b7723b8c2bfc0b900d)

src/apps/frontend/pages/authentication/phone-login/phone-login-form.hook.ts
saisudan2003: I moved the `.toString()` call from the assignment line (` const formattedPhoneNumber = parsedPhoneNumber .getNationalNumber().toString();`)  to the sendOTP() call because `parsedPhoneNumber.getNationalNumber()` was triggering a TypeScript warning:
`"Object is possibly 'undefined'".`

Instead of using optional chaining (`?.toString()`) or type assertions, I added a runtime check `(!formattedPhoneNumber)` to ensure it's safely defined before converting it to a string.

src/apps/backend/modules/task/rest-api/task-controller.ts
```diff
@@ -53,10 +53,10 @@ export class TaskController {
   );
 
   getTasks = applicationController(async (req: Request, res: Response) => {
-    const page = +req.query.page;
-    const size = +req.query.size;
+    const page = req.query.page ? +req.query.page : 1; // Default to page 1 if not provided
+    const size = req.query.size ? +req.query.size : 10; // Default to size 10 if not provided
```
govind-io, line 57: Unnecessary comments for something which is evident by code itself.

src/apps/frontend/vendor/inspectlet.ts
```diff
@@ -1,19 +1,19 @@
...
+  window.__insp.push(['wid', window.Config.inspectLetKey]);
   const ldinsp = (): void => {
     if (typeof window.__inspld !== 'undefined') return;
     window.__inspld = 1;
     const insp = document.createElement('script');
     insp.type = 'text/javascript';
     insp.async = true;
     insp.id = 'inspsync';
-    insp.src = `${document.location.protocol === 'https:' ? 'https' : 'http'}://cdn.inspectlet.com/inspectlet.js?wid=${String(window.Config.inspectletKey)}&r=${Math.floor(new Date().getTime() / 3600000)}`;
+    insp.src = `${document.location.protocol === 'https:' ? 'https' : 'http'}://cdn.inspectlet.com/inspectlet.js?wid=${String(window.Config.inspectLetKey)}&r=${Math.floor(new Date().getTime() / 3600000)}`;
```
govind-io, line 11: Why is the`inspectLetKey` case changed? This is not something related to typescript,

src/apps/frontend/types/globals.d.ts
```diff
@@ -3,12 +3,15 @@ export {};
 declare global {
   const Config: {
     env: string;
-    inspectletKey: string;
+    inspectLetKey: string;
```
govind-io, line 6: any reason for changing this name?

src/apps/frontend/vendor/inspectlet.ts
saisudan2003: This change is part of an older conversation which was marked as resolved - [link](https://github.com/jalantechnologies/boilerplate-mern/pull/198#discussion_r1944114262)

src/apps/frontend/types/globals.d.ts
saisudan2003: This change is part of an older conversation which was marked as resolved - [link](https://github.com/jalantechnologies/boilerplate-mern/pull/198#discussion_r1944114262)

src/apps/backend/modules/task/rest-api/task-controller.ts
saisudan2003: Resolved via commit [ec1bd5d](https://github.com/jalantechnologies/boilerplate-mern/pull/198/commits/ec1bd5d1a0f713e195456b3d9f589802b77c39c5)

test/spec/task/task.spec.ts
```diff
@@ -194,7 +194,9 @@ describe('Task API', () => {
       expect(res.status).to.eq(204);
 
       const updatedToken = await TaskRepository.findById(task.id);
-      expect(updatedToken.active).to.be.false;
+
+      expect(updatedToken).to.exist;
```
govind-io, line 198: This is a change not related to typescript and rather updating

src/apps/backend/modules/authentication/internals/password-reset-token/password-reset-token-writer.ts
```diff
@@ -25,12 +25,20 @@ export default class PasswordResetTokenWriter {
...
-    return PasswordResetTokenRepository.findByIdAndUpdate(
+    const updatedToken = await PasswordResetTokenRepository.findByIdAndUpdate(
       passwordResetTokenId,
       {
         isUsed: true,
       },
       { new: true }
     );
+
+    if (!updatedToken) {
```
govind-io, line 36: This is not a change related to typescript but a functionality change, but this is done in order to prevent the type error. We are not in control that the token id being passed is always correct, neither there is a way to statically verify that so in order to prevent the type error instead of doing inference we have added this check.

src/apps/backend/modules/task/rest-api/task-controller.ts
```diff
@@ -53,10 +53,10 @@ export class TaskController {
   );
 
   getTasks = applicationController(async (req: Request, res: Response) => {
-    const page = +req.query.page;
-    const size = +req.query.size;
+    const page = req.query.page ? +req.query.page : 1;
+    const size = req.query.size ? +req.query.size : 10;
```
govind-io, line 57: Setting up default values, previous assumption was that `page` and `size` will always be defined in `req.query` and was not handling for a scenario where this is not passed.

src/apps/frontend/pages/tasks/index.tsx
```diff
@@ -32,11 +32,12 @@ const Tasks: React.FC = () => {
     <div className="mx-auto max-w-5xl">
       <VerticalStackLayout gap={7}>
         <HeadingMedium>TaskList</HeadingMedium>
-        <TaskHeader />
+        <TaskHeader onError={onError} />
```
govind-io, line 35: This is changed here as the `TaskHeader` `onError` prop is made `required` now as previously one of child component of the `TaskHeader` needed that prop but `TaskHeader` was not passing any value to it.

test/spec/authentication/account-auth-middleware.spec.ts
```diff
@@ -64,7 +64,7 @@ describe('accessAuthMiddleware', () => {
               authorization: `Bearer ${accessToken.token}`,
             },
           } as unknown as Request,
-          undefined,
+          {} as unknown as Response,
```
govind-io, line 67: This works as `accessAuthMiddleware` is simply ignoring the value of `_res` but to make it compatible with express we have to define the type of `_res` to`Response` so we can't make it `Nullable` there otherwise `express` would not let us set it as a middleware citing that it's a wrong type.

src/apps/frontend/contexts/account.provider.tsx
```diff
@@ -1,24 +1,25 @@
...
 };
 
-const AccountContext = createContext<AccountContextType | null>(null);
+const AccountContext = createContext<Nullable<AccountContextType>>(null);
 
 const accountService = new AccountService();
 
 export const useAccountContext = (): AccountContextType =>
-  useContext(AccountContext);
+  useContext(AccountContext) as AccountContextType;
```
govind-io, line 22: This `as` assertion is needed as the value of `useContext(AccountContext)` is `Nullable` as well.

govind-io (approved):

PR #199, opened 2025-02-05
feat: add support for running cyclic dependency check

### LINKED ISSUE #63: (feat) Add support for running cyclic dependency check
Add a new check to pipeline which would check for cyclic dependencies. It should fail (just like test/lint) if any cyclic dependencies were found.

sn1f3rt:
## Description

Add support for running cycling dependency check on PR preview and production push via [dependency-cruiser](https://www.npmjs.com/package/dependency-cruiser) | Resolves #63 

An example of logs when this test fails: https://github.com/jalantechnologies/boilerplate-mern/actions/runs/13159811196/job/36725637271

### Manual Tests

Ran the cyclic dependency checks and resolved each one of them (there were total of 4 when the checks were added)


src/apps/backend/common/account-password-reset-common.ts
jjalan, line 1: This is not Ok - feels like we are using common as an escape goat to patch the problem.

sn1f3rt: Okay I commented before the edit, @saisudan2003 can we look into this?

saisudan2003: Sure will look into it.

.dependency-cruiser.js
```diff
@@ -0,0 +1,397 @@
+/** @type {import('dependency-cruiser').IConfiguration} */
+module.exports = {
```
jjalan, line 2: Is this copied from somewhere - if yes, can we add a reference to that link in this file as comment?

src/apps/backend/modules/account/types.ts
```diff
@@ -58,21 +58,13 @@ export type GetAccountParams = {
...
 export type UpdateAccountDetailsParams = {
   accountId: string;
   firstName: string;
   lastName: string;
 };
 
-export type UpdateAccountParams =
-  | UpdateAccountDetailsParams
-  | ResetPasswordParams;
+export type UpdateAccountParams = UpdateAccountDetailsParams;
```
govind-io, line 67: Any reason that we need these two types to be exactly same? Can we not just use `UpdateAccountDetailsParams` everywhere or `UpdateAccountParams` whatever suits? 

Can we remove this duplication?

govind-io (approved):

src/apps/backend/modules/account/types.ts
sn1f3rt: Updated via https://github.com/jalantechnologies/boilerplate-mern/pull/199/commits/71369cd44c88c72a778f14d270b413f3d4405d01

govind-io (approved):

package.json
```diff
@@ -18,6 +18,7 @@
     "lint": "eslint .",
     "lint:md": "remark .",
     "lint:fix": "eslint --fix .",
+    "check:cycles": "npx depcruise src",
```
jjalan, line 21: nit `check: cyclic-dependencies`

src/apps/backend/modules/documentation/internals/document-generator-util.ts
```diff
@@ -114,7 +114,7 @@ export default class DocumentGeneratorUtil {
       const routerFileContent = fs.readFileSync(routerFilePath, 'utf8');
 
       const routeRegex = new RegExp(
-        `router\\.${route.method.toLowerCase()}\\(['"]${route.routerPath}['"],\\s*ctrl\\.(\\w+)\\)`,
+        `router\\.${route.method.toLowerCase()}\\(\\s*['"]${route.routerPath}['"]\\s*,\\s*ctrl\\.(\\w+)\\s*,?\\s*\\)`,
```
jjalan, line 117: Can you explain the changes and why they are needed?

src/apps/backend/modules/password-reset-token/rest-api/password-reset-token-router.ts
```diff
@@ -8,5 +8,9 @@ export default class PasswordResetTokenRouter extends ApplicationRouter {
     const ctrl = new PasswordResetTokenController();
 
     router.post('/password-reset-tokens', ctrl.createPasswordResetToken);
+    router.patch(
+      '/password-reset-tokens/validate',
```
jjalan, line 12: This is not a CRUD endpoint, CRUD dictates POST, PATCH, DELETE, GET, GET :id on the entity.

src/apps/backend/modules/password-reset-token/types.ts
```diff
@@ -19,9 +19,16 @@ export type CreatePasswordResetTokenParams = {
...
+export type ValidatePasswordResetTokenAndResetPasswordParams = {
+  accountId: string;
+  newPassword: string;
+  token: string;
+};
+
 export enum PasswordResetTokenErrorCode {
   PASSWORD_RESET_TOKEN_NOT_FOUND = 'PASSWORD_RESET_TOKEN_ERR_01',
   PASSWORD_RESET_EMAIL_NOT_ENABLED_FOR_THE_ENVIRONMENT = 'PASSWORD_RESET_TOKEN_ERR_02',
+  BAD_REQUEST = 'ACCOUNT_ERR_05',
```
jjalan, line 31: Most probably should be `PASSWORD_RESET_TOKEN_ERR_03`

jjalan (commented):
I think we are trying to patch a solution here to move updating password into `password-reset-token` module by exposing a `/validate` route. The problem is its not CRUD operation.

Have we considered the approach discussed [here](https://teams.microsoft.com/l/message/19:08e4a50edf634281a15e27fa7ca99fef@thread.tacv2/1739172074618?tenantId=79836b2a-53cc-4854-81b4-ba2d7c9f2726&groupId=76c0380b-15c1-4a18-8308-24c4abc83332&parentMessageId=1739171449082&teamName=Kernel&channelName=Platform%20(Team)&createdTime=1739172074618&ngc=true)?

src/apps/backend/modules/password-reset-token/types.ts
sn1f3rt: Updated via 2134c42395ba030a6c5274856b1fb106d043a3d8

src/apps/backend/modules/documentation/internals/document-generator-util.ts
sn1f3rt: This is for supporting multiline routes, basically what was happening is that after the files were being formatted the line was too long to fit in one line so it was being expanded to multiline. But the document generator was not being able to read the route since the regex only supported single line routes. That's why this change was made. I think we should preserve it for future-proofing.

src/apps/backend/modules/password-reset-token/rest-api/password-reset-token-router.ts
sn1f3rt: Reverted to CRUD functionality via 335e7963a3d04f1342c804361e77e95f92eaef4c

At the moment, this might not look entirely convincing, mostly because the patch is being called on the password reset token route to update the account's password, which is misleading. However, there is an open issue at the moment #217, to restructure the folders and when that is done this functionality will be separated into auth and account and this particular functionality will move back into account as per your suggestion in the teams chat link you mentioned, and hence we can proceed with the current implementation for now.

PR #201, opened 2025-02-06
[WIP] Added Badge Reports to README

PunithKosana:
## Description
_This Pull Request addressess on adding badge reports for Build, Tests, Coverage to README file._
- Build Badge: Indicates the latest GitHub Actions build status.
- Test Badge: Shows the test workflow result.
- Coverage Badge: Displays overall test coverage percentage.

## Database schema changes
_NA._

-NA

### Manual test cases run
- Verified the paths of the Build, Tests, Coverage reports are correctly mentioned.
- Verified the url's on browser on generating the reports.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Boilerplate project for NodeJS, React based projects in TypeScript. This README 
 
 | Build | Tests | Code Coverage |
 |--------|--------|--------|
-| [![Build Status](https://img.shields.io/github/actions/workflow/status/jalantechnologies/boilerplate-mern/.github/workflows/preview_on_pr_update.yml?branch=main&label=Build&job=build)] | [![Test Status](https://img.shields.io/github/actions/workflow/status/jalantechnologies/boilerplate-mern/.github/workflows/preview_on_pr_update.yml?branch=main&label=Tests&job=checks)] | [![Code Coverage](https://img.shields.io/badge/Code%20Coverage-90%25-success?style=flat)] |
+| ![Build](https://img.shields.io/github/actions/workflow/status/jalantechnologies/boilerplate-mern/preview_on_pr_update.yml)| [![Tests](https://img.shields.io/github/actions/workflow/status/jalantechnologies/boilerplate-mern/preview_on_pr_update.yml)] | [![Code Coverage](https://img.shields.io/badge/Code%20Coverage-90%25-success?style=flat)] |
 
 
 ## Table of Contents
--- a/README.md
+++ b/README.md
@@ -2,6 +2,13 @@
 
 Boilerplate project for NodeJS, React based projects in TypeScript. This README documents whatever steps are necessary to get your application up and running.
 
+## Badge Reports
+
+| Build | Tests | Code Coverage |
+|--------|--------|--------|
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,6 +2,13 @@
 
 Boilerplate project for NodeJS, React based projects in TypeScript. This README documents whatever steps are necessary to get your application up and running.
 
+## Badge Reports
+
+| Build | Tests | Code Coverage |
+|--------|--------|--------|
+| [![Build Status](https://img.shields.io/github/actions/workflow/status/jalantechnologies/boilerplate-mern/.github/workflows/preview_on_pr_update.yml?branch=main&label=Build&job=build)] | [![Test Status](https://img.shields.io/github/actions/workflow/status/jalantechnologies/boilerplate-mern/.github/workflows/preview_on_pr_update.yml?branch=main&label=Tests&job=checks)] | [![Code Coverage](https://img.shields.io/badge/Code%20Coverage-90%25-success?style=flat)] |
```
jjalan, line 9: This always display the health of the main branch. This works for main branch.

1. But say if I have raised a PR and go to README for that branch, it should display health of that branch, is that possible?
2. When we clone a new app, how would this work - it sounds like it will always show boilerplate health, not the app health. If it cannot be automated, do we need to update README

We would also need a demo video from preview app showing this work end to end in PR description

README.md
```diff
@@ -2,6 +2,13 @@
 
 Boilerplate project for NodeJS, React based projects in TypeScript. This README documents whatever steps are necessary to get your application up and running.
 
+## Badge Reports
+
+| Build | Tests | Code Coverage |
+|--------|--------|--------|
+| ![Build](https://img.shields.io/github/actions/workflow/status/jalantechnologies/boilerplate-mern/preview_on_pr_update.yml)| [![Tests](https://img.shields.io/github/actions/workflow/status/jalantechnologies/boilerplate-mern/preview_on_pr_update.yml)] | [![Code Coverage](https://img.shields.io/badge/Code%20Coverage-90%25-success?style=flat)] |
```
govind-io, line 9: The code coverage is manually set to 90% it should dynamically come.

govind-io: But in that case the code coverage badge is of no use, we should remove it and implement it once we have a proper solution for it.

govind-io: Also right now both for build and Tests we are rendering the exact same badge.

govind-io: Also not mentioning the branch name in Build or Test Badge uses the default branch name which is main branch, it does not reflects the status of the specific branch. If it's not possible to dynamically inject the branch name let's keep it as the status of the main branch it self for now.

govind-io:
@PunithKosana , we should try rendering a github page as our readme file and in that page we should be able to use javascript to dynamically get the current branch's name and insert it to the shield io badge url.

README.md
PunithKosana: > But in that case the code coverage badge is of no use, we should remove it and implement it once we have a proper solution for it.

Sure. Will update that

PunithKosana: > Also right now both for build and Tests we are rendering the exact same badge.

It is same. Because as we are triggering the pipeline from outside the repo i.e `github-ci` . We don't have seperate jobs and checks for that. Also, the pipeline runs on the `preview_on_pr_update.yml` as it is same for the both jobs(Build,Test)

PunithKosana: > Also not mentioning the branch name in Build or Test Badge uses the default branch name which is main branch, it does not reflects the status of the specific branch. If it's not possible to dynamically inject the branch name let's keep it as the status of the main branch it self for now.

Ok. Will add that

PR #204, opened 2025-02-07
fix: docker dev build

### LINKED ISSUE #203: fix: docker build in watch mode does fails on macOS
Build in watch mode using `docker compose -f docker-compose.dev.yml up` fails on macOS.  

Repro steps:

- Clone boilerplate-mern on macOS
- Run - `docker compose -f docker-compose.dev.yml up`
- The build fails with `sh: cross-env command not found`

Image

sn1f3rt:
## Description

Resolves #203

## Manual Test Cases

Previously the docker build would exit as under:

image

After applying this fix, I tried running the docker build command, and it worked without any issues
 
image

I've confirmed the same with @saisudan2003, this issue was specific to macOS.

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -12,7 +12,7 @@ services:
     environment:
       NODE_ENV: 'development'
       NODE_APP_INSTANCE: 'docker'
-    command: 'npm run serve'
+    command: /bin/sh -c "npm install && npm run serve"
     depends_on:
       - db
     ports:


docker-compose.dev.yml
```diff
@@ -12,7 +12,7 @@ services:
     environment:
       NODE_ENV: 'development'
       NODE_APP_INSTANCE: 'docker'
-    command: 'npm run serve'
+    command: /bin/sh -c "npm install && npm run serve"
```
jjalan, line 15: Can you add test cases section and help us understand better that this fix indeed fixed the problem with screenshots and/or other ways that can be easily verified.

sn1f3rt: Added the issue and resolution screenshots in both the issue and this PR.

sn1f3rt: Another thing to note is that this is how it's done in frm-bp and it works fine on mac: https://github.com/jalantechnologies/rflask-boilerplate/blob/main/docker-compose.dev.yml#L14

jjalan (approved):

PR #206, opened 2025-02-10
fix(webpack): wait for backend to get ready

### LINKED ISSUE #60: (fix) The webpack hot reload server should wait for backend to get ready
- When attempted to run `npm run serve`, the script opens up the url automatically on default browser of the dev machine
- When frontend build is finished, the page loads but with broken assets (open up network tab, there will be connection refused requests made by the browser).
- This happens because assets are actually served by the backend, and till it's not, it won't start listening to any incoming connections. Hence, the connection refused error.
- We need to find a way to delay thescript to emit frontend build till backend is ready.

sn1f3rt:
## Description

Webpack hot reload server should wait for backend to get ready | Resolves #60 

### Manual test cases run

Ran `npm run serve`, and ensured that the webpack server starts only after the backend server has finished loading. 

image


.prettierrc.yml
```diff
@@ -1,2 +1,2 @@
 singleQuote: true
-trailingComma: all
+trailingComma: es5
```
jjalan, line 2: The change seems unrelated to this PR

src/apps/frontend/webpack.dev.js
```diff
@@ -20,21 +53,35 @@ const config = {
...
+    },
+    setupMiddlewares: (middlewares, devServer) => {
+      if (!devServer) {
+        throw new Error('Webpack DevServer is not defined');
+      }
+
+      waitForBackend(devServerAPIProxyPort, () => {
+        console.log('Starting Webpack Dev Server...');
+        if (devServerOpen) {
+          open(`http://localhost:${devServerPort}`);
```
jjalan, line 80: Like `API_SERVER_URL`, perhaps `http://localhost:${devServerPort}` store in a constant.

src/apps/frontend/webpack.dev.js
```diff
@@ -1,13 +1,46 @@
...
 
 const devServerOpen = process.env.WEBPACK_DEV_DISABLE_OPEN !== 'true';
 const devServerPort = 3000;
 const devServerAPIProxyPort = 8080;
 
+const API_SERVER_URL = `http://localhost:${devServerAPIProxyPort}`;
 const DOCUMENTATION_PAGE_ROUTE = '/api/documentation';
 
+function waitForBackend(port, callback) {
+  const retryInterval = 2000;
```
jjalan, line 15: Move this outside function, all caps. Also use unit in variable name to make it readable - ex `RETRY_INTERVAL_IN_MS` or something similar

src/apps/frontend/webpack.dev.js
```diff
@@ -1,13 +1,46 @@
...
+          console.log(`Backend is ready on port ${port}`);
+          callback();
+        } else {
+          console.log(`Waiting for backend on port ${port}...`);
+          setTimeout(check, retryInterval);
+        }
+      }
+    );
+
+    req.on('error', () => {
```
jjalan, line 31: When does this happen?

Is this saying webpack dev server errored, it will continuously keep trying?

src/apps/frontend/webpack.dev.js
```diff
@@ -1,13 +1,46 @@
...
 const devServerPort = 3000;
 const devServerAPIProxyPort = 8080;
 
+const API_SERVER_URL = `http://localhost:${devServerAPIProxyPort}`;
 const DOCUMENTATION_PAGE_ROUTE = '/api/documentation';
 
+function waitForBackend(port, callback) {
+  const retryInterval = 2000;
+
+  function check() {
```
jjalan, line 17: I see there are logs in each if, wondering if all are necessary from a dev standpoint. We want to make sure its not very verbose when the app starts, almost feels like it should say nothing and silently do the right thing, but when it cannot, it should console.error

src/apps/frontend/webpack.dev.js
```diff
@@ -20,21 +53,35 @@ const config = {
...
           // To place the documentation page on the respective route, the proxy should be bypassed for this route.
           if (req.url === DOCUMENTATION_PAGE_ROUTE) {
             return '/index.html';
           }
         },
       },
-      '/assets': `http://localhost:${devServerAPIProxyPort}`,
+      '/assets': API_SERVER_URL,
+    },
+    setupMiddlewares: (middlewares, devServer) => {
```
jjalan, line 72: I am not familiar with this, can you elaborate on how this works a bit?

.prettierrc.yml
sn1f3rt: Well unfortunately, it is, setting trailing comma to all results in a prettier eslint conflict and I wouldn't have been able to commit the files otherwise. I'm attaching a screenshot below. 

image

The solution to this would be to upgrade the versions of eslint and prettier and ensure compatibility but that is up to another PR, for now we should only allow trailing commas per the es5 syntax as that is what is being parsed by the version of eslint here.

src/apps/frontend/webpack.dev.js
sn1f3rt: Updated via https://github.com/jalantechnologies/boilerplate-mern/pull/206/commits/e43ade583548f9ecda6f93274b9c5c523c55c334

src/apps/frontend/webpack.dev.js
sn1f3rt: Updated via https://github.com/jalantechnologies/boilerplate-mern/pull/206/commits/e43ade583548f9ecda6f93274b9c5c523c55c334

src/apps/frontend/webpack.dev.js
sn1f3rt: Basically in the case where the backend server hasn't started accepting connections yet, and yes it keeps retrying, should I add a `RETRY_LIMIT`?

src/apps/frontend/webpack.dev.js
sn1f3rt: I created a demo video illustrating the changes -

https://github.com/user-attachments/assets/1384e6d1-9f87-4487-87dd-a96d1272fa50

src/apps/frontend/webpack.dev.js
sn1f3rt: Sure, this is for custom logic handling before the requests are forwarded to webpack, and here we are waiting for the backend before opening the browser which webpack would have done prematurely before.

src/apps/frontend/webpack.dev.js
jjalan: Backend not retrying in 2 sec appears quite a lot of time, when I saw this - I was wondering is there anything wrong? Retrying logs typically suggest something is wrong.

src/apps/frontend/webpack.dev.js
jjalan: >Basically in the case where the backend server hasn't started accepting connections yet

Why would webpack dev server error in that case?

Killg0d (approved):

src/apps/frontend/webpack.dev.js
sn1f3rt: > > Basically in the case where the backend server hasn't started accepting connections yet
> 
> Why would webpack dev server error in that case?

Because it would result in connection failures if the backend hasn't started accepting any requests on the port yet.

src/apps/frontend/webpack.dev.js
sn1f3rt: That is because I individually started the frontend only to illustrate the changes ... in practice when the backend server is already up it would result in an instant connection, or if it is the docker dev build where both the backend and frontend are being built and started simultaneously, it would wait for a couple of seconds at max before the connection is established.

jjalan (approved):
I am Ok with these changes, please merge once @govind-io  also +1

govind-io (approved):

PR #218, opened 2025-03-04
Add comment feature to FE and implement create comment API in BE.

Developer01ash:
## FE: 
- Created a separate page (task-comments.tsx) to manage task comments.
- Integrated the new page into the task section for easy navigation.

## BE: 
- Implemented a new endpoint to create comments on tasks.
**API end point**- http://localhost:3000/api/comments
**API Body**-
{
    "taskId": "67c6c24fb319e066d479095c",
    "content": "Hello test"
}

## Database schema changes
**Task===>**
`export const TaskDbSchema = new Schema<TaskDB>(
  {
    active: {
      type: Boolean,
      required: true,
      default: true,
    },
    account: {
      type: Schema.Types.ObjectId,
      ref: 'Account',
      index: true,
      required: true,
    },
    description: {
      type: String,
      required: true,
    },
    title: {
      type: String,
      required: true,
    },
    sharedWith: [
      {
        type: Schema.Types.ObjectId,
        ref: 'Account',
        required: true,
      },
    ],
    comments: [
      {
        type: Schema.Types.ObjectId,
        ref: 'Comment',
      },
    ],
  },
  {
    collection: 'tasks',
    timestamps: {
      createdAt: 'createdAt',
      updatedAt: 'updatedAt',
    },
  }
);`
**Comment==>**
`export const CommentDbSchema = new Schema<CommentDB>(
  {
    content: {
      type: String,
      required: true,
    },
    account: {
      type: Schema.Types.ObjectId,
      ref: 'Account',
      required: true,
    },
    task: {
      type: Schema.Types.ObjectId,
      ref: 'Task',
      required: true,
    },
  },
  {
    collection: 'comments',
    timestamps: {
      createdAt: 'createdAt',
      updatedAt: 'updatedAt',
    },
  }
);`

PR #220, opened 2025-03-05
WIP: Implement API for Sharing Tasks with Users

Developer01ash:
This pull request implements the API functionality for sharing tasks with users. The new endpoint allows users to share tasks with other users, storing the association between tasks and shared users in a new ShareTask collection. This API ensures that the task is shared only with valid users and tasks, and it responds with the necessary information, including populated references to the task and the user.
And PR adds unit tests for the **Task API** and **Comment API** endpoints using the Given/When/Then naming convention. The following test cases have been added:

**Key Features:**
```
- Added an API endpoint to share tasks with a user.
- Validated the existence of both tasks and users before sharing.
- Introduced a new ShareTask model to represent shared tasks.
- Populated the task and sharedWith fields in the response to include full task and user data.
- Ensured proper error handling in case of invalid task or user.
```

## Database schema changes
This PR introduces changes to the database schema to support the new feature of sharing tasks between users. Specifically, a new ShareTask model has been created, and a new collection shared_tasks will store the associations between tasks and users.

## Tests
**Automated test cases added**
```
POST /tasks/share
- Given: A task and another account.
     - When: The task is shared with the other account.
     - Then: It should respond with the shared task details.
- Given: A task and a non-existent account ID.
      - When: The task is shared with the non-existent account ID.
      - Then: It should respond with a 404 status and an error message.
- Given: A non-existent task ID and another account.
      - When: The task is shared with the non-existent task ID.
      - Then: It should respond with a 404 status and an error message.
```

```
POST /comments
- Given: Valid parameters for creating a comment.
      - When: The endpoint is called.
      - Then: It should return the created comment.
```


src/apps/backend/app.ts
```diff
@@ -84,6 +86,14 @@ export default class App {
         serverInstance: new TaskServer(),
         rootFolderPath: path.join(__dirname, 'modules/task'),
       },
+      {
+        serverInstance: new CommentServer(),
+        rootFolderPath: path.join(__dirname, 'modules/comment'),
+      },
+      {
+        serverInstance: new ShareTaskServer(),
```
jjalan, line 94: Sharing a task should belong to task module.

The scope of module is somewhat subjective, but the guidance is to group logical functionality in a same module.

src/apps/backend/modules/comment/rest-api/comment-serializer.ts
```diff
@@ -0,0 +1,8 @@
+import { Comment } from '../types';
+
+export const serializeCommentAsJSON = (comment: Comment): unknown => ({
+  id: comment.id,
+  account: comment.account,
+  taskId: comment.taskId,
+  content: comment.content,
```
jjalan, line 7: nit: order alphabetically

src/apps/backend/modules/comment/internal/store/comment-db.ts
```diff
@@ -0,0 +1,34 @@
...
+export interface CommentDB {
+  _id: Types.ObjectId;
+  account: Types.ObjectId;
+  content: string;
+  task: Types.ObjectId;
+}
+
+export const CommentDbSchema = new Schema<CommentDB>(
+  {
+    content: {
```
jjalan, line 12: nit: order alphabetically here and other places

Also, would someone be able to add emoji and other type of non UTF8 content type here?

src/apps/backend/modules/comment/types.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export enum CommentErrorCode {
+  NOT_FOUND = 'COMMENT_ERR_01',
+}
+
+export class CommentNotFoundError extends ApplicationError {
+  code: CommentErrorCode;
+
+  constructor(taskId: string) {
+    super(`Comment with taskId ${taskId} not found.`);
```
jjalan, line 25: Error should offer solution instead of blaming or just surfacing the problem. Please google "how to write good error messages"

src/apps/backend/modules/share-task/internal/share-task-util.ts
```diff
@@ -0,0 +1,12 @@
+import { ShareTask } from '../types';
+
+import { ShareTaskDB } from './store/share-task-db';
+
+export default class ShareTaskUtil {
+  public static convertTaskDBToTask(shareTaskDb: ShareTaskDB): ShareTask {
+    const sareTask = new ShareTask();
```
jjalan, line 7: nit: spelling mistake

src/apps/backend/modules/share-task/rest-api/share-task-server.ts
```diff
@@ -0,0 +1,12 @@
+import { ApplicationServer } from '../../application';
+
+import ShareTaskRouter from './share-task-router';
+
+export default class ShareTaskServer extends ApplicationServer {
+  configure(): void {
+    const { server } = this;
+    const router = new ShareTaskRouter();
+
+    server.use('/share-with', router.router);
```
jjalan, line 10: This is not a CRUD API.

CRUD API should work with entities such as user, etc. `share-with` is a noun. This is not Ok.

You probably have to expose `share-requests` or something which would queue up a request to share a task with a user.

jjalan (commented):
Left cursory comments, did not review PR in detail.

Best to write spec for comment, sharing functionality before moving forward

src/apps/backend/app.ts
Developer01ash: As per the [PR comment](https://github.com/jalantechnologies/boilerplate-mern/pull/218#discussion_r1979541776) on the task module, I thought it was necessary to create a new module for sharing tasks, similar to how I created the 'post comment' feature in the new task module.

src/apps/backend/modules/share-task/rest-api/share-task-server.ts
Developer01ash: Thank you for the feedback! I see your point about the difference between a CRUD API and the purpose of this functionality. I'll adjust the approach and implement a 'share-request' feature that queues the request to share a task with a user, instead of treating it as a direct CRUD operation. I'll update the design accordingly and submit a new version for review.

src/apps/backend/modules/comment/internal/store/comment-db.ts
Developer01ash: String type in Mongoose already supports UTF-8 encoding, which includes emojis and other special characters.

Developer01ash:
The comment APIs for adding, retrieving, editing, and deleting have been integrated into the frontend. will continuing work on refining the user interface.

Developer01ash:
Comment on task - post, view, edit and delete done at FE.

src/apps/backend/modules/account/account-service.ts
```diff
@@ -106,4 +106,8 @@ export default class AccountService {
 
     return AccountWriter.deleteAccountById(accountId);
   }
+
+  public static async getAllAccounts(params: string): Promise<Account[]> {
```
jjalan, line 110: I am not sure the rationale behind exposing this method but this is *highly* undesirable.

An application in production can have 1000's of accounts and loading of them in memory is not good. Furthermore, we never want to expose all accounts from a privacy point of view as well

jjalan, line 110: `params: string`: The name of the arg is ambiguous and not sure what it `params` refer to

src/apps/backend/modules/comment/internal/comment-reader.ts
```diff
@@ -0,0 +1,34 @@
...
+import CommentUtil from './comment-util';
+import CommentRepository from './store/comment-repository';
+
+export default class CommentReader {
+  public static async getCommentsForTask(
+    params: GetCommentsParams
+  ): Promise<Comment[]> {
+    const commentsDb = await CommentRepository.find({
+      task: params.taskId,
+    }).populate('account', 'firstName lastName username');
```
jjalan, line 17: This is not OK.

By directly inspecting the `account` database, you are violating the boundary definition aka now the comment module is taking dependency on something very internal to `account` module. Anything that is not exposed by AccountService or its types.ts file is considered internal and its not Ok to take dependency on it.

src/apps/backend/modules/comment/internal/comment-writer.ts
```diff
@@ -0,0 +1,59 @@
...
+export default class CommentWriter {
+  public static async createComment(
+    params: CreateCommentParams
+  ): Promise<Comment> {
+    const createdComment = await CommentRepository.create({
+      account: params.accountId,
+      content: params.content,
+      task: params.taskId,
+    });
+    const populatedComment = await createdComment.populate(
```
jjalan, line 21: same comment as above.

src/apps/backend/modules/comment/internal/comment-writer.ts
```diff
@@ -0,0 +1,59 @@
...
+        _id: params.commentId,
+      },
+      {
+        $set: {
+          content: params.content,
+        },
+      },
+      { new: true }
+    );
+    const populatedComment = await updatedComment.populate(
```
jjalan, line 43: same comment as above

src/apps/backend/modules/comment/internal/store/comment-db.ts
```diff
@@ -0,0 +1,36 @@
...
+      type: String,
+      required: true,
+      trim: true,
+      minlength: 1,
+    },
+    task: {
+      type: Schema.Types.ObjectId,
+      ref: 'Task',
+      required: true,
```
jjalan, line 26: Do we need to index this?

Can you check CommentReader file and see if we need to create other combined indexes?

src/apps/backend/modules/comment/rest-api/comment-router.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export default class CommentRouter extends ApplicationRouter {
+  configure(): void {
+    const { router } = this;
+    const ctrl = new CommentController();
+
+    router.use(accessAuthMiddleware);
+
+    router.post('/', ctrl.createComment);
+    router.get('/:taskId', ctrl.getComments);
```
jjalan, line 14: This is not CRUD API.

GET `/comments?task_id=?` is the right paradigm.

src/apps/backend/modules/account/account-service.ts
Developer01ash: Should I implement a search functionality that allows querying accounts based on specific criteria, such as email ID, to avoid loading all accounts into memory and exposing sensitive information?

PR #222, opened 2025-03-10
chore: upgrade TypeScript version

saisudan2003:
## Description
Upgraded TypeScript to `5.8.2`

### Changes made:

- Upgraded Cypress to `14.1.0` for TypeScript compatibility
    - Issue: Cypress `13.2.0` was incompatible with TypeScript `5.8.2`, causing compilation errors.
    - Fix: Upgraded Cypress to `14.1.0`.

- Installed `@types/node@22.13.10` to resolve TypeScript type issues
  - Issue: Missing `@types/node `caused type resolution errors, particularly in module.d.ts due to stricter overload signature checks in `TypeScript 5.8.2`.
  - Fix: Installed @types/node@22.13.10.
  - Reference : [Link](https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/70562)

- Updated `intervalId` type from` NodeJS.Timer` to `NodeJS.Timeout`
  - Issue: Installing `@types/node` enforced correct type definitions, revealing that `NodeJS.Timer` is not a valid type. The correct type for `setInterval` and `setTimeout` return values is `NodeJS.Timeout`.
  - Fix: Replaced `NodeJS.Timer` with `NodeJS.Timeout` to align with Node.js type definitions.

### DIFF
--- a/src/apps/frontend/pages/profile/settings/index.tsx
+++ b/src/apps/frontend/pages/profile/settings/index.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import toast from 'react-hot-toast';
 import { useNavigate } from 'react-router-dom';
 
@@ -18,7 +18,7 @@ const ProfileSettings = () => {
   const navigate = useNavigate();
 
   const [isDeleteAccountModalOpen, setIsDeleteAccountModalOpen] =
-    React.useState(false);
+    useState(false);
 
   const handleLogout = () => {
     logout();
--- a/src/apps/frontend/pages/profile/settings/index.tsx
+++ b/src/apps/frontend/pages/profile/settings/index.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React from 'react';
 import toast from 'react-hot-toast';
 import { useNavigate } from 'react-router-dom';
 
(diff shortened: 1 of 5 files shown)


sn1f3rt (approved):

tsconfig.json
```diff
@@ -1,6 +1,8 @@
 {
   "compilerOptions": {
     "target": "es6",
+    "types": ["node", "mocha", "chai"],
+    "skipLibCheck": true,
```
jjalan, line 5: Remove.

tsconfig.json
```diff
@@ -1,6 +1,8 @@
 {
   "compilerOptions": {
     "target": "es6",
+    "types": ["node", "mocha", "chai"],
```
jjalan, line 4: Ideally, this should not be needed.

src/apps/frontend/pages/profile/settings/index.tsx
```diff
@@ -18,7 +18,7 @@ const ProfileSettings = () => {
   const navigate = useNavigate();
 
   const [isDeleteAccountModalOpen, setIsDeleteAccountModalOpen] =
-    useState(false);
+    React.useState(false);
```
jjalan, line 21: This sounds like a bandaid solution. We should look into the right way to import, using `React.` seems undesirable.

jjalan (approved):
Can you update the PR description based on the final changes?

PR #255, opened 2025-11-13
Add security context hardening to Docker and Kubernetes deployments

rehanshaikh1207:
## Description
This PR addresses security vulnerabilities and network policy gaps in the node-react-template. The problems solved include running containers as root, writable root filesystems, and improper network policy labeling which could lead to privilege escalation, container compromise, and incorrect network isolation.

video Walkthrough: https://www.loom.com/share/3b3013c008ae4f3e954d2abd4bae3b8b

## Changes

- **Dockerfile**: Added non-root user appuser (UID 10001) with proper directory ownership

- **Kubernetes Deployments**: Added pod and container security contexts with read-only root filesystem

- **Security Features**: Disabled privilege escalation, dropped all Linux capabilities, added writable volume mounts

- **Network Policy**: Added component labels for proper network policy matching

### Manual test cases run

_Test 1: Confirm vulnerability exists when running as root in preview deployment_

- created a test PR with the original configuration (before changes).
 
- removed the root files essential for the application which results in this error shown below

- This demonstrates the security vulnerability where a compromised container has high privileges and risks affecting the host system or other containers as it could allow attackers to execute root commands on the host or affect other containers, since container isolation is not absolute.
Screenshot 2025-11-13 180744
Screenshot 2025-11-13 180641

_Manual Test 2: Verify User and Security Context_

- Verified the pod runs as user appuser (UID/GID 10001).

- Confirmed pod-level security context enforces runAsNonRoot, sets filesystem group, and enforces runtime default seccomp.

- Checked container-level security context disables privilege escalation, drops all capabilities, and sets a read-only root filesystem.

- Ensured pod labels match expected deployment metadata.
Screenshot 2025-11-13 165717

### DIFF
--- a/lib/kube/preview/deployment.yaml
+++ b/lib/kube/preview/deployment.yaml
@@ -114,3 +114,4 @@ spec:
           emptyDir: {}            # Temporary storage for log files
         - name: system-tmp
           emptyDir: {}            # Temporary storage for system files
+


bettrhq:
@codex

PR #257, opened 2025-11-13
Pin all GitHub Actions to commit SHAs and enable Dependablebot scaning

Namra2511:
**Why these changes were required**

Dependabot has been added to automatically scan for updates in all pinned GitHub Action SHAs.
This ensures that whenever a new release (commit SHA) is published for any action used in our workflows, Dependabot will raise an automated pull request with the updated SHA reference.

Using immutable commit SHAs instead of floating version tags (like @v4 or @v5) eliminates the risk of unverified code changes being pulled into the CI/CD pipeline.This approach improves supply-chain security and prevents unexpected behavior caused by re-tagged or updated versions.

**What changes were made**

- Added a Dependabot configuration file (.github/dependabot.yml) which runs from the repository’s default branch.
- Updated all GitHub workflow files to replace version tags (@vX) with their corresponding commit SHAs for both first-party and third-party actions.
- Previously, during testing, Dependabot was generating around 90 PRs across npm, pip, and GitHub Actions updates.
With grouping enabled, it now raises a single PR per ecosystem, reducing CI noise and simplifying review.

**Manual testing done**
**1) Added the Dependabot configuration to the main branch of a forked repository and manually triggered Dependabot from insight.**

image
image

**Loom video** : https://www.loom.com/share/72f378a0c68945b2a25604b717dd0239

### DIFF
--- a/.github/workflows/permanent_preview.yml
+++ b/.github/workflows/permanent_preview.yml
@@ -52,3 +52,4 @@ jobs:
           do_access_token: ${{ secrets.DO_ACCESS_TOKEN }}
           do_cluster_id: ${{ vars.DO_CLUSTER_ID }}
           doppler_token: ${{ secrets.DOPPLER_PREVIEW_TOKEN }}
+


.github/dependabot.yml
```diff
@@ -0,0 +1,28 @@
...
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 5
+    groups:
+      gha-all:
+        patterns:
+          - "*"
+    labels:
+      - "dependencies"
+      - "security"
```
joeldsouza28: add EOL

joeldsouza28 (approved):

Namra2511:
> I see 08c6903cd8c0fde910a37f88322edcfb5dd907a8 repeated many times. If we have to change it in future, it makes it error prone in case someone forgets to change it at one place but not others. I am not sure if this is true for other SHA's.
> 
> Is there a way to create a variable and use it across?

GitHub Actions only allows two types of values in the uses: field:
(1) a literal version tag (e.g., @v4) or
(2) a literal commit SHA.

Both of these must be written as static literal strings.
GitHub does not allow variables, YAML anchors, environment variables, or any kind of indirect reference to be used inside uses:. This is a security restriction to prevent dynamic resolution of actions at runtime.

Because of this limitation, even though we are using commit SHAs instead of tags, we still cannot centralize them into a variable. Each uses: line must reference the exact SHA explicitly.

